PR #3, opened 2018-08-08
Feature/multi location artifact build

halfhorst:
Works but haven't produced an actual artifact with it.

The qsub response is logged but it isn't shown. The overall job monitoring could be made more pretty and informative.


src/scripts/build_artifact.py
```diff
@@ -0,0 +1,196 @@
...
+    simulation = InteractiveContext(simulation_config, components, plugin_manager)
+    simulation.data.start_processing(simulation.component_manager, output_path,
+                                     [simulation.configuration.input_data.location],
+                                     incremental=not args.from_scratch)
+    simulation.setup()
+    simulation.data.end_processing()
+
+
+def update_configuration(specification_arg, location_arg, output_root_arg, configuration):
+    """Update the simulation configuration artifact output path and location with
```
collijk: There's a lot of duplication here as well.  Think about how you might split this up into pieces that work together.  Some suggestions:

1. A `get_location(location_arg, config)` function.  This gives you back the location in the config or the location arg.  Either way, you can always update the configuration with it. Or raise an error.
2. A `contains_location` function or variable.  `'input_data' in configuration and 'location' in configuration.input_data and configuration.input_data.location` is an arcane magic statement that works around limitations of a data structure, but that's not likely to be clear to someone reading this code.
3. A `get_output_path(specification_name, output_root, location_arg)` function.  

This should collect all the relevant logic into decoupled pieces and let you write something like:
```
configuration.input_data.location = get_location(location_arg, configuration)
configuration.artifact.path = get_output_path(specification_name, output_path, location_arg)
```
actually, that's it.  The end.

src/scripts/build_artifact.py
```diff
@@ -0,0 +1,196 @@
...
+    split_response = response.split()
+    # qsub response can be funky but JID should directly follow "Your job".
+    # job arrays say "job-array", regular jobs say "job"
+    job_ind = 0
+    if 'job' in split_response:
+        job_ind = split_response.index('job')
+    elif 'job-array' in split_response:
+        job_ind = split_response.index('job-array')
+    if not job_ind:
+        raise OSError("Unexpected response from qsub: " + response)
```
collijk: Same thing about the OSError.

src/scripts/build_artifact.py
```diff
@@ -0,0 +1,196 @@
...
+from vivarium.interface.interactive import InteractiveContext
+
+
+_log = logging.getLogger(__name__)
+
+
+@click.command()
+@click.argument('simulation_configuration', nargs=1)
+@click.argument('project', nargs=1)
+@click.argument('locations', nargs=-1)
```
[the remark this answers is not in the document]
halfhorst: if you mean locations, it is because only args are compatible with variadic parameters (nargs=-1). "For options, only a fixed number of arguments is supported"

halfhorst:
Documenting some of my click decisions:
*arguments: required, positional
*options: optional
*locations are actually optional but variadic, which requires argument
*arguments don't allow help, options do.
*General help is taken from docstring so it should be formatted as we want the user to see, and arguments should be documented there as well. I read a few man pages to get a flavor for how these are worded.
*I generally accept defaults, nargs=1 and type=click.STRING

collijk:
Since you went through the trouble of writing all the docstrings, let's talk through how to supply type information for arguments and returns before we merge this in.

src/scripts/build_artifact.py
```diff
@@ -0,0 +1,155 @@
...
+    and a configuration file"""
+
+    configuration_path = pathlib.Path(configuration_arg)
+
+    if output_root_arg:
+        output_base = pathlib.Path(output_root_arg).resolve()
+    else:
+        output_base = pathlib.Path('/share') / 'scratch' / 'users' / getpass.getuser() / 'vivarium_artifacts'
+
+    if location_arg:
```
halfhorst: @collijk: we clobber the output in the config no matter what. We either take user provided or use a default. Should we be adding location info no matter what? I.e. either what they provided OR introspect the config and pull out the location?

halfhorst:
@collijk Updated with numpy docstrings and also quite a few pep8 changes, mostly linewrap

collijk:
This looks great!

PR #4, opened 2018-08-09
Refactor/core tests

halfhorst:
Remove a mock from core.py and mock vivarium_gbd_access in test_core.py

### DIFF
--- a/tests/test_core.py
+++ b/tests/test_core.py
@@ -216,9 +216,7 @@ def test_get_draws_bad_args(cause_list, risk_list, locations):
 
 def test_get_relative_risk(mocker):
     gbd_mock = mocker.patch("vivarium_inputs.core.gbd")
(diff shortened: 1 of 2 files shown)


tests/test_core.py
```diff
@@ -215,6 +215,7 @@ def test_get_draws_bad_args(cause_list, risk_list, locations):
 
 
 def test_get_relative_risk(mocker):
+    gbd_mock = mocker.patch("vivarium_inputs.core.gbd")
     age_group_mock = mocker.patch("vivarium_inputs.core.gbd.get_age_group_id")
```
collijk: I don't actually think you need the `age_group_mock = mocker.patch("vivarium_inputs.core.gbd.get_age_group_id")` call.  Mocks automatically generate attributes as called.  Just replace all the `age_group_mock`s with `gbd_mock.get_age_group_id`

PR #6, opened 2018-08-17
Split data artifact into a package

collijk:


collijk:
@michaella21 Yep.  This was just moving things around.  I submitted it separately so we don't have full file diffs again when there are actual changes.

PR #8, opened 2018-08-24
Hotfix/build artifact default

halfhorst:
Some logic was flipped around a bit, it became apparent when i tried to build some artifacts. Specifying a path where there was no data, i.e. I want to make a new artifact, resulted in an error unless --from-scratch was included. 

The original argument was from-scratch and seemed intended to indicate the user wanted to overwrite data, but it was saved into a config slot that was for appending, and what hdf's touch method did with it didn't really follow what the argument meant. Additionally, I misinterpreted what store_true did as a default. 

Additionally there is now a default cluster project set, proj_cost_effect.

Originally there was more here, but now it is mostly a name change. There is a corresponding PR to vivarium_public_health for hdf's touch method which interprets the append configuration a bit more.

### DIFF
--- a/src/vivarium_inputs/data_artifact/cli.py
+++ b/src/vivarium_inputs/data_artifact/cli.py
@@ -19,7 +19,7 @@ from vivarium.config_tree import ConfigTree
                 readable=True))
 @click.argument('locations', nargs=-1)
 @click.option('--project', default='proj_cost_effect')
-@click.option('--output_root', type=click.Path(file_okay=False, writable=True),
+@click.option('--output-root', type=click.Path(file_okay=False, writable=True),
               help="Directory to save artifact result in")
 @click.option('--append', type=click.BOOL, default=False,
               help="Preserve existing artifact and append to it")
@@ -130,7 +130,7 @@ def _build_artifact():
     parser = argparse.ArgumentParser()
     parser.add_argument('model_specification', type=str,
                         help="path to a model_specification file")
-    parser.add_argument('--output_root', type=str, required=False,
+    parser.add_argument('--output-root', type=str, required=False,
                         help="directory to save artifact to. "
                              "Overwrites model_specification file")
     parser.add_argument('--location', type=str, required=False,
--- a/src/vivarium_inputs/data_artifact/cli.py
+++ b/src/vivarium_inputs/data_artifact/cli.py
@@ -17,13 +17,14 @@ from vivarium.config_tree import ConfigTree
 @click.command()
 @click.argument('model_specification', type=click.Path(dir_okay=False,
                 readable=True))
-@click.argument('project')
 @click.argument('locations', nargs=-1)
+@click.option('--project', default='proj_cost_effect')
 @click.option('--output_root', type=click.Path(file_okay=False, writable=True),
               help="Directory to save artifact result in")
 @click.option('--from_scratch', type=click.BOOL, default=True,
               help="Do not reuse any data in the artifact, if any exists")
 def build_artifact(model_specification, project, locations,
+def build_artifact(model_specification, locations, project,
                    output_root, from_scratch):
     """
     build_artifact is a program for building data artifacts from a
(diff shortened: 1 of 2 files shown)


src/vivarium_inputs/data_artifact/cli.py
```diff
@@ -129,15 +130,14 @@ def _build_artifact():
...
                         help="directory to save artifact to. "
                              "Overwrites model_specification file")
     parser.add_argument('--location', type=str, required=False,
                         help="location to get data for. "
                              "Overwrites model_specification file")
-    parser.add_argument('--from_scratch', '-s', action="store_true",
-                        help="Do not reuse any data in the artifact, "
-                             "if any exists")
+    parser.add_argument('--append', action="store_true",
+                        help="Append to an artifact if it exists")
```
kfwilson: Maybe make this help message the same as the message for the same argument in build_artifact? I find that one a little clearer that it will break if the artifact doesn't exist. This suggests that it might only apply if the artifact exists and will just pass through and create it if the artifact doesn't exist - at least that was my first thought when I read it

src/vivarium_inputs/data_artifact/cli.py
```diff
@@ -17,23 +17,24 @@
...
 @click.argument('locations', nargs=-1)
-@click.option('--output_root', type=click.Path(file_okay=False, writable=True),
+@click.option('--project', default='proj_cost_effect')
+@click.option('--output-root', type=click.Path(file_okay=False, writable=True),
               help="Directory to save artifact result in")
-@click.option('--from_scratch', type=click.BOOL, default=True,
-              help="Do not reuse any data in the artifact, if any exists")
-def build_artifact(model_specification, project, locations,
-                   output_root, from_scratch):
+@click.option('--append', type=click.BOOL, default=False,
```
collijk: This is a much better default.

src/vivarium_inputs/data_artifact/cli.py
```diff
@@ -17,23 +17,24 @@
 @click.command()
 @click.argument('model_specification', type=click.Path(dir_okay=False,
                 readable=True))
-@click.argument('project')
 @click.argument('locations', nargs=-1)
-@click.option('--output_root', type=click.Path(file_okay=False, writable=True),
+@click.option('--project', default='proj_cost_effect')
```
collijk: Let's also add some short flags for the options.  `-P` for projcect, '-o' for output_root, '-a' for append.

michaella21: just checking, I don't think most people have an access to `proj_cost_effect`, pretty sure only we have an access to it. maybe it's better to keep it default as before?

src/vivarium_inputs/data_artifact/cli.py
```diff
@@ -17,23 +17,24 @@
...
-    overwrite the configuration file.
+    SIMULATION_CONFIGURATION file. The work is offloaded to the cluster
+    under the "proj_cost_effect" project unless otherwise specified. 
+    Multiple, optional LOCATIONS can be provided to overwrite the configuration
+    file.
 
     Any artifact.path specified in the configuration file is guaranteed to
     be overwritten either by the optional output_root or a predetermined path
-    based on user: /ihme/scratch/{user}/vivarium_artifacts
+    based on user: /ihme/scratch/users/{user}/vivarium_artifacts
```
kfwilson: I think it's a little confusing that this default path (which isn't mentioned in the doc for _build_artifact) will break if the user hasn't already created it. It seems like if we're giving users a default and not telling always telling them about it, we should ensure it exists?

src/vivarium_inputs/data_artifact/cli.py
collijk: If they don't have access to `proj_cost_effect`, they likely shouldn't be building artifacts.  Or they'll know which project they do have access to and be able to submit it as an option.

src/vivarium_inputs/data_artifact/cli.py
collijk: We should actually make the path if it doesn't exist. 

The generation of directories should happen in this script.  The failure point doesn't occur til we call hdf.touch.  The normal touch command also expects intermediary directories to exist, so we want to be consistent.

halfhorst: Kate and I discussed, she is going to address this as part of her logging PR so she can be sure the directory exists and can log to it

halfhorst:
This is updated. More changes to come from Kate

PR #10, opened 2018-08-29
filter out cause_ids which are not in gbd_mapping; relative_risk

michaella21:
quick fix so that we can have smoking_prevalence in data artifact

### DIFF
--- a/src/vivarium_inputs/core.py
+++ b/src/vivarium_inputs/core.py
@@ -422,7 +422,7 @@ def _get_relative_risk(entities, location_ids):
         measure_data.loc[:, draw_cols] = 1/measure_data.loc[:, draw_cols]
         measure_data = _handle_coverage_gap_data(entities, measure_data, 1)
 
-    cause_ids = [causes.__getitem__(c).gbd_id for c in causes.__slots__]
+    cause_ids = [c.gbd_id for c in causes]
     valid_cause_ids = [c for c in measure_data['cause_id'].unique() if c in cause_ids]
     measure_data = measure_data[measure_data['cause_id'].isin(valid_cause_ids)]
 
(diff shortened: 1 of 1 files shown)


src/vivarium_inputs/core.py
```diff
@@ -423,6 +428,9 @@ def _get_relative_risk(entities, location_ids):
         measure_data.loc[:, draw_cols] = 1/measure_data.loc[:, draw_cols]
         measure_data = _handle_coverage_gap_data(entities, measure_data, 1)
 
+    valid_cause_ids = [c for c in measure_data['cause_id'].unique() if c in _get_cause_ids()]
```
collijk: I think you're doing a lot more work here than you need to.  
```
valid_cause_ids = [cause_id for cause_id in measure_data['cause_id'].unique() if cause_id in [c.gbd_id for c in causes]]
```
Does that break somehow?  I'm not sure why you have the helper function where you're mucking about in the slots.

michaella21: it didn't really work well in one line when I tried. I will fix it a little later, since it's not the most urgent one right now

halfhorst:
This seems like it will do some of the same things that Kate's logging PR will do wrt cli.py ! You each converged on the same UI improvements. I don't think it's a big deal, you can work it out on merge

src/vivarium_inputs/core.py
```diff
@@ -423,6 +422,10 @@ def _get_relative_risk(entities, location_ids):
         measure_data.loc[:, draw_cols] = 1/measure_data.loc[:, draw_cols]
         measure_data = _handle_coverage_gap_data(entities, measure_data, 1)
 
+    cause_ids = [c.gbd_id for c in causes]
+    valid_cause_ids = [c for c in measure_data['cause_id'].unique() if c in cause_ids]
```
collijk: You don't even need to do this.  
```
measure_data = measure_data[measure_data['cause_id'].isin([c.gbd_id for c in causes]))
``` 
should do the trick.

collijk: Or in two lines for clarity:
```
most_detailed_causes = measure_data['cause_id'].isin([c.gbd_id for c in causes]
measure_data = measure_data[most_detailed_causes]
```

PR #11, opened 2018-09-06
Bugfix/artifact builder filters

michaella21:
This is a quick fix for `ArtifactBuilder` which misses filter terms(draw and location) and fails in building an artifact during the setup phase. ( basically, I added the filter terms symmetrical to the `ArtifactManager` in vivarium_public_health. With this fix, users should be able to build an artifact as expected.

### DIFF
--- a/src/vivarium_inputs/data_artifact/builder.py
+++ b/src/vivarium_inputs/data_artifact/builder.py
@@ -1,8 +1,9 @@
 from datetime import datetime
 import logging
 from typing import Collection, Any
+import pandas as pd
 
-from vivarium_public_health.dataset_manager import EntityKey, Artifact, hdf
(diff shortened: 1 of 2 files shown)


src/vivarium_inputs/data_artifact/builder.py
```diff
@@ -18,8 +19,9 @@ def setup(self, builder):
         append = builder.configuration.input_data.append_to_artifact
         hdf.touch(path, append)
 
-        self.artifact = Artifact(path)
         self.location = builder.configuration.input_data.location
+        self.draw = builder.configuration.input_data.input_draw_number
```
collijk: Draw doesn't need to be stored as an attribute on the builder.

PR #12, opened 2018-09-14
Feature/auxiliary coverage gap data

michaella21:
This PR is all about pulling the new coverage gap data into our system. Since it's the first time that we have the coverage gap data other than `low_measles_vaccine_coverage_first_dose` (which is still from gbd), I can't really think of any clean and clever way to do it other than having all special case functions. Regarding the age_group_id 22, I considered to drop the age column in that case. However too many places in our system expects to have exposure/relative_risks, etc have the age columns. So I just assigned the midpoint of (0,125). If there's a better way to do it, please let me know. With current fixes, the artifact builder should be able to load the coverage gap exposure/rr/paf to the data artifact.

### DIFF
--- a/src/vivarium_inputs/utilities.py
+++ b/src/vivarium_inputs/utilities.py
@@ -168,7 +168,7 @@ def normalize_for_simulation(df):
             df_f = df.copy()
             df_m['sex'] = 'Male'
             df_f['sex'] = 'Female'
-            df = pd.concat([df_m, df_f])
+            df = pd.concat([df_m, df_f], ignore_index=True)
         else:
             df["sex"] = df.sex_id.map({1: "Male", 2: "Female", 3: "Both"}).astype(
                 pd.api.types.CategoricalDtype(categories=["Male", "Female", "Both"], ordered=False))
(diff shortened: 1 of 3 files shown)


src/vivarium_inputs/utilities.py
```diff
@@ -163,10 +163,18 @@ def normalize_for_simulation(df):
...
     """
     if "sex_id" in df:
-        df["sex"] = df.sex_id.map({1: "Male", 2: "Female", 3: "Both"}).astype(
-            pd.api.types.CategoricalDtype(categories=["Male", "Female", "Both"], ordered=False))
+        if set(df["sex_id"]) == {3}:
+            df_m = df.copy()
+            df_f = df.copy()
+            df_m['sex'] = 'Male'
+            df_f['sex'] = 'Female'
+            df = pd.concat([df_m, df_f])
```
halfhorst: the index here is duplicated. Should it be reset?

michaella21:
I tried to restructure as much as possible according to what we discussed. Regarding the last function in the loader, I considered moving it to core, but just left it since core still works based on the ids (risk/cause etc) and loaders convert number ids to the actual name. Added/modified some standardization functions for all age_groups/both_sexes/any missing indexes.

src/vivarium_inputs/utilities.py
michaella21: yeah, these kinds of simple duplication, I change it to `ignore_index=True`

PR #16, opened 2018-09-25
Fix incidence calculation

halfhorst:
This will convert the incidence we get from get_draws (total incidence) to the incidence we want/expect, the hazard rate.

### DIFF
--- a/src/vivarium_inputs/core.py
+++ b/src/vivarium_inputs/core.py
@@ -309,12 +309,22 @@ def _get_incidence(entities, location_ids):
     correct_measure = measure_data['measure_id'] == name_measure_map['incidence']
     correct_years = measure_data['year_id'].isin(gbd.get_estimation_years(gbd.GBD_ROUND_ID))
 
-    pop_incidence = measure_data[correct_measure & correct_years]
-    pop_prevalence = _get_prevalence(entities, location_ids)
+    # get_draws returns "total incidence." We want hazard incidence.
+    # scale by the number of people not sick to convert.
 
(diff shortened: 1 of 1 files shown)


src/vivarium_inputs/core.py
```diff
@@ -308,7 +308,23 @@ def _get_incidence(entities, location_ids):
...
+    incidence = measure_data[correct_measure & correct_years]
+    prevalence = _get_prevalence(entities, location_ids)
+
+    draw_columns = [col for col in incidence.columns if col.startswith('draw_')]
+
+    incidence.set_index(['age_group_id', 'location_id', 'cause_id',
+                             'sex_id', 'year_id'], inplace=True)
+    prevalence.set_index(['age_group_id', 'location_id', 'cause_id',
+                              'sex_id', 'year_id'], inplace=True)
+
```
michaella21: I'm not 100% sure, but pandas set_index probably doesn't stay until you re-define like `incidence=incidence.set_index...`

halfhorst: it does only because I have inplace=True there

michaella21: ahhhhh I missed that

collijk:
We need to make sure this doesn't break anything.  Can you run the smoke tests in ceam_experiments?

halfhorst:
will do

halfhorst:
breaks on 'cause_id' sometimes, don't know what guarantees I have around what columns will be around. Need to be able to line up the quotient between the incidence and prevalence dataframes. Looking in to it

PR #17, opened 2018-09-27
Hardcoded super gross fixes to pafs

collijk:


src/vivarium_inputs/data_artifact/builder.py
```diff
@@ -49,4 +49,5 @@ def process(self, entity_key: EntityKey) -> None:
 def _worker(entity_key: EntityKey, location: str, modeled_causes: Collection[str], artifact: Artifact) -> None:
     for measure, data in loader(entity_key, location, modeled_causes, all_measures=True):
         key = entity_key.with_measure(measure)
-        artifact.write(key, data)
+        if key not in artifact:
```
kfwilson: when would the data already be in the artifact? Or are we doing this because the private method can technically be called from somewhere else where we don't have the guarantee that the data hasn't already been processed?

src/vivarium_inputs/data_artifact/loaders.py
```diff
@@ -439,8 +439,11 @@ def _handle_coverage_gap_rr_paf(data):
     data = data.rename(columns={'cause_id': 'cause', 'rei_id': 'risk_factor'})
     if data['cause'].dropna().unique().size > 0:
         for cid in data['cause'].dropna().unique():
-            data['cause'] = data['cause'].applyl(lambda c: CAUSE_BY_ID[c].name if c == cid else c)
+            data['cause'] = data['cause'].apply(lambda c: CAUSE_BY_ID[c].name if c == cid else c)
     if data['risk_factor'].dropna().unique().size > 0:
         for rid in data['risk_factor'].dropna().unique():
-            data['risk_factor'] = data['risk_factor'].apply(lambda r: RISK_BY_ID[r].name if r == rid else r)
+            if rid == 318:  # measles
+                data.loc[data['risk_factor'] == rid, 'risk_factor'] = 'low_measles_vaccine_coverage_first_dose'
```
kfwilson: maybe use coverage_gaps.low_measles_vaccine_coverage_first_dose.name? I don't know how often (if ever) we foresee coverage gap names changing, but if that's something that might happen, it'd be easier just to use name here.

src/vivarium_inputs/data_artifact/builder.py
collijk: This happens when you resume a build.  It's a result of our strategy of bringing in all measures associated with an entity the first time we see an entity.  If a build fails or gets stopped while it's in the middle of writing all the measures, only some of them will be present in the artifact when we restart.

The simulation then runs for a while, skips a bunch of entity measures because they're in the half built artifact, finds a measure that isn't in the artifact, and then tries to write all the measures into the artifact again.

src/vivarium_inputs/data_artifact/loaders.py
collijk: The names aren't here, just the ids.  A dramatically more serious revision is necessary to make something reasonable here, but we just need to be able to build artifacts for the diarrhea model first.

src/vivarium_inputs/data_artifact/loaders.py
```diff
@@ -439,8 +439,11 @@ def _handle_coverage_gap_rr_paf(data):
     data = data.rename(columns={'cause_id': 'cause', 'rei_id': 'risk_factor'})
     if data['cause'].dropna().unique().size > 0:
         for cid in data['cause'].dropna().unique():
-            data['cause'] = data['cause'].applyl(lambda c: CAUSE_BY_ID[c].name if c == cid else c)
+            data['cause'] = data['cause'].apply(lambda c: CAUSE_BY_ID[c].name if c == cid else c)
```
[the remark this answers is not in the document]
collijk: :+1:

src/vivarium_inputs/core.py
```diff
@@ -540,13 +542,19 @@ def _compute_paf_for_special_cases(affected_entity: Union[Cause, Risk], entity:
...
         temp_result['risk_id'] = risk_id
         temp_result['measure_id'] = 3
         paf.append(temp_result.reset_index())
 
     paf_data = pd.concat(paf)
+
+    if entity == coverage_gaps.low_measles_vaccine_coverage_first_dose:
+        paf_data['rei_id'] = entity.gbd_id
+        paf_data['coverage_gap_id'] = entity.gbd_id
+
```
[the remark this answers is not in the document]
collijk: This system is very fragile right now and we don't have tests around it.  Be very careful if you're going to change something.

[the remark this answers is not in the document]
michaella21: yep, I will build/run model with measles coverage gap to test it after my change as well as smoke tests

PR #19, opened 2018-10-01
Bugfix/artifact builder not properly get modeled causes

michaella21:
Quick fixes to have an artifact builder correctly load the coverage gap data. Since coverage gap data can both come from gbd and auxiliary data source, it is messier and does not have very standardized form yet. There should be more fixes to clean up as we get more coverage gap data coming in. For now, I let it go as it is. **But current any function (get exposure/relative_risk/paf) which may use coverage gap in core is very fragile. Please be careful when you have to change/drop any column name to make a change in those functions**

### DIFF
--- a/src/vivarium_inputs/data_artifact/cli.py
+++ b/src/vivarium_inputs/data_artifact/cli.py
@@ -163,7 +163,7 @@ def _build_artifact():
     _setup_logging(args.output_root, args.verbose, args.location,
                    args.model_specification, args.append)
 
-    formatted_location = args.location.replace('_', ' ')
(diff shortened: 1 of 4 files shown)


src/vivarium_inputs/data_artifact/cli.py
```diff
@@ -163,7 +163,7 @@ def _build_artifact():
     _setup_logging(args.output_root, args.verbose, args.location,
                    args.model_specification, args.append)
 
-    formatted_location = args.location.replace('_', ' ')
+    formatted_location = args.location.replace('_', ' ') if args.location else args.location
```
halfhorst: :smile:

PR #21, opened 2018-10-10
Feature/add common covariates

halfhorst:
This adds support for a list of covariates that Abie indicated he would want to use for simulations some time going forward. Because these and the existing covariates are all formatted the same, I made a change to retain the age column as a general rule. All smoke tests pass except Sanofi

### DIFF
--- a/src/vivarium_inputs/data_artifact/loaders.py
+++ b/src/vivarium_inputs/data_artifact/loaders.py
@@ -226,14 +226,14 @@ def get_covariate_data(covariate, measure, location, _):
         expected_columns = ["location", "mean_value", "lower_value", "upper_value",
                             "sex_id", "year_id", "age_group_id"]
 
-        if expected_columns not in data.columns():
+        if expected_columns not in data.columns:
             raise NotImplementedError(f"Unsupported covariate {covariate.name}. It does not not "
                                        f"contain the columns {expected_columns}.")
 
         data = data[expected_columns]
         data = get_age_group_midpoint_from_age_group_id(data)
     else:
-        raise NotImplementedError((f"Unknown measure or unsupported {measure} for ")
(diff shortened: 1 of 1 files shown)


src/vivarium_inputs/data_artifact/loaders.py
```diff
@@ -222,20 +222,26 @@ def get_population_data(_, measure, location, __):
 
 def get_covariate_data(covariate, measure, location, _):
     if measure == "estimate":
-        data = core.get_covariate_estimates([covariate], [location])
-        if covariate.name == "age_specific_fertility_rate":
-            columns = ["location", "mean_value", "lower_value", "upper_value", "sex_id", "year_id", "age_group_id"]
+        if covariate.name in ["age_specific_fertility_rate", "live_births_by_sex",
```
collijk: Instead of explicitly white listing things, I think it might be worthwhile to look for particular data shapes.  Let people pull anything with the columns ["location", "mean_value", "lower_value", "upper_value",  "sex_id", "year_id", "age_group_id"] and then mark things for special consideration.

collijk: Also, does this still work for age_specific fertility rate?.  That should only have values for sex_id 2.

halfhorst: So you mean pull first and just check what columns are there ?

collijk: Yeah, pull, make sure columns are there and make sure they have values we know how to interpret.  Maybe raise a warning for, e.g., only one sex_id or only a subset of age_group_ids present.

halfhorst:
I broke things being more permissive with covariates. Standby

halfhorst:
I added back a function call I left out that normalizes column names. Also added support for age_group _22 (which doesn't actually break things, but results in an age column of NaN!). It's ready to look at

PR #22, opened 2018-10-17
Refactor/core

michaella21:
This PR is mainly changing the input arguments type from `Iterable` to a single entity for the most functions. Most of changes is changing `[entities]` to a single entity and any other necessary changes from it. 

The only exception is for `coverage_gap` PAF/RR and where it is loaded in loader. Not every coverage gap from auxiliary folder has relative risk while the loader still call for the PAF/RR of every coverage gap. To avoid the error, I let it catch the `NotADirectoryError`. There should be a better way to handle this. But in either way we need to check this since users are actively adding new data to aux folder lately.

### DIFF
--- a/src/vivarium_inputs/core.py
+++ b/src/vivarium_inputs/core.py
@@ -83,15 +83,15 @@ class DuplicateDataError(DataError):
     pass
 
 
-def get_draws(entities: Sequence[ModelableEntity], measures: Iterable[str],
+def get_draws(entity: ModelableEntity, measures: Iterable[str],
               locations: Iterable[str]) -> pd.DataFrame:
     """Gets draw level gbd data for each specified measure and entity.
 
     Parameters
     ----------
-    entities:
-        A list of data containers from the `gbd_mapping` package. The entities must all be the same
-        type (e.g. all `gbd_mapping.Cause` objects or all `gbd_mapping.Risk` objects, etc.
+    entity:
+        A data container from the `gbd_mapping` package.
+        (e.g. `gbd_mapping.Cause` object or `gbd_mapping.Risk` object, etc.
     measures:
         A list of the GBD measures requested for the provided entities.
     locations:
@@ -127,13 +127,13 @@ def get_draws(entities: Sequence[ModelableEntity], measures: Iterable[str],
             handler, id_columns = measure_handlers[measure]
         except KeyError:
             raise InvalidQueryError(f'No functions are available to pull data for measure {measure}')
-        measure_data = handler(entities, location_ids)
+        measure_data = handler(entity, location_ids)
         measure_data['measure'] = measure
         id_cols |= id_columns
         data.append(measure_data)
     data = pd.concat(data)
 
-    id_cols |= _get_additional_id_columns(data, entities)
+    id_cols |= _get_additional_id_columns(data, entity)
     key_columns = ['measure']
     for column in ['year_id', 'sex_id', 'age_group_id', 'location_id']:
         if column in data:
@@ -152,7 +152,7 @@ def get_draws(entities: Sequence[ModelableEntity], measures: Iterable[str],
 
 
 # TODO: Move to utilities.py
-def _get_ids_for_measure(entities: Sequence[ModelableEntity], measure: str) -> List:
+def _get_ids_for_measure(entity: ModelableEntity, measure: str) -> List:
     """Selects the appropriate gbd id type for each entity and measure pair.
 
     Parameters
@@ -189,30 +189,25 @@ def _get_ids_for_measure(entities: Sequence[ModelableEntity], measure: str) -> L
         'cost': ((HealthcareEntity, HealthTechnology), 'cost'),
     }
 
-    if not all([isinstance(e, type(entities[0])) for e in entities]):
-        raise InvalidQueryError("All entities must be of the same type")
     if measure not in measure_types.keys():
         raise InvalidQueryError(f"You've requested an invalid measure: {measure}")
 
     valid_types, id_attr = measure_types[measure]
-    out = []
-    for entity in entities:
-        if isinstance(entity, valid_types):
-            value = entity
-            for sub_id in id_attr.split('.'):
-                if value[sub_id] is not UNKNOWN:
-                    value = value[sub_id]
-                else:
-                    raise InvalidQueryError(f"Entity {entity.name} has no data for measure '{measure}'")
-            out.append(value)
-        else:
-            raise InvalidQueryError(f"Entity {entity.name} has no data for measure '{measure}'")
 
-    return out
+    if isinstance(entity, valid_types):
+        value = entity
+        for sub_id in id_attr.split('.'):
+            if value[sub_id] is not UNKNOWN:
+                value = value[sub_id]
+            else:
+                raise InvalidQueryError(f"Entity {entity.name} has no data for measure '{measure}'")
+        return [value]
+    else:
+        raise InvalidQueryError(f"Entity {entity.name} has no data for measure '{measure}'")
 
 
 # TODO: Move to utilities.py
-def _get_additional_id_columns(data, entities):
+def _get_additional_id_columns(data, entity):
     id_column_map = {
         Cause: 'cause_id',
         Sequela: 'sequela_id',
@@ -224,8 +219,8 @@ def _get_additional_id_columns(data, entities):
         HealthTechnology: 'health_technology',
     }
     out = set()
-    out.add(id_column_map[type(entities[0])])
-    if isinstance(entities[0], CoverageGap) and data['measure'].all() in ['relative_risk']:
+    out.add(id_column_map[type(entity)])
+    if isinstance(entity, CoverageGap) and data['measure'].all() in ['relative_risk']:
         out.add('risk_id')
     out |= set(data.columns) & set(id_column_map.values())
     return out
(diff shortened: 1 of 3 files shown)


src/vivarium_inputs/data_artifact/loaders.py
```diff
@@ -434,12 +434,12 @@ def _get_coverage_gap_relative_risk(coverage_gap, location):
...
 def _get_coverage_gap_population_attributable_fraction(coverage_gap, location):
-    data = core.get_draws([coverage_gap], ["population_attributable_fraction"], [location])
-    if data.empty:
-        data = None
-    else:
+    try:
+        data = core.get_draws([coverage_gap], ["population_attributable_fraction"], [location])
         data = _handle_coverage_gap_rr_paf(data)
         data = data[["year", "location", "cause", "risk_factor", "sex", "draw", "value", "age"]]
+    except core.InvalidQueryError:
```
halfhorst: Should this actually be raised so people know they asked for invalid data?

michaella21: I had the exact same question for some other parts of loaders (remission). It's for the entity that doesn't always have the all the measures. Some coverage gaps don't have Paf/rr, some causes don't have remission, but we can't really differentiate them case by case. So decided just to pass the none data instead. If this missing data is still called during the setup, it should break but otherwise it shouldn't break the process of building an artifact. There should be a better way to handle this, but I haven't figured that out. :(

src/vivarium_inputs/core.py
```diff
@@ -189,30 +189,25 @@ def _get_ids_for_measure(entities: Sequence[ModelableEntity], measure: str) -> L
...
+    if isinstance(entity, valid_types):
+        value = entity
+        for sub_id in id_attr.split('.'):
+            if value[sub_id] is not UNKNOWN:
+                value = value[sub_id]
+            else:
+                raise InvalidQueryError(f"Entity {entity.name} has no data for measure '{measure}'")
+        return [value]
+    else:
+        raise InvalidQueryError(f"Entity {entity.name} has no data for measure '{measure}'")
```
kfwilson: Should this be a different error message than above? This case isn't just that this particular entity doesn't have data for the measure, it's that the user shouldn't be asking for this measure for this entity type at all

src/vivarium_inputs/core.py
```diff
@@ -467,47 +456,38 @@ def _pull_rr_data_from_gbd(measure_ids):
...
         measure_data = measure_data[most_detailed_causes]
 
-    elif isinstance(entities[0], CoverageGap):
-        df = []
+    elif isinstance(entity, CoverageGap):
         SPECIAL = [coverage_gaps.low_measles_vaccine_coverage_first_dose]
-        special_cases = set(SPECIAL).intersection(set(entities))
-        if special_cases:
-            measure_data = _pull_rr_data_from_gbd([i for i in measure_ids if i is not None])
+        special_cases = set(SPECIAL).intersection(set(entity))
```
kfwilson: You don't really need special_cases anymore. You're already checking if entity is in SPECIAL so you can just pass entity instead of special_cases to _handle_special_coverage_gap_data on line 475 below

src/vivarium_inputs/core.py
```diff
@@ -529,8 +509,12 @@ def _compute_paf_for_special_cases(affected_entity: Union[Cause, Risk], entity:
...
     :param affected_entity: which entity is affected by this entity
     """
-    ex = _get_exposure([entity], location_ids)
-    rr = _get_relative_risk([entity], location_ids)
+    ex = _get_exposure(entity, location_ids)
+    rr = _get_relative_risk(entity, location_ids)
+
+    # coverage gap data from auxiliary folder may not have rr but will still call for paf
+    if ex.empty or rr.empty:
+        raise InvalidQueryError(f'You requested the PAF for the entity that may not have necessary data')
```
kfwilson: Can we explicitly tell the user what data they're missing?

src/vivarium_inputs/core.py
```diff
@@ -647,71 +631,68 @@ def handle_exposure_from_gbd(measure_data):
...
-                measure_data = measure_data[measure_data.location_id.isin(location_ids)]
-                measure_data['coverage_gap'] = entity.name
-                del measure_data['measure']
-                df.append(measure_data)
+        else:
+            measure_data = gbd.get_auxiliary_data('exposure', 'coverage_gap', entity.name)
+            measure_data = measure_data[measure_data.location_id.isin(location_ids)]
+            measure_data['coverage_gap'] = entity.name
+            del measure_data['measure']
+            df.append(measure_data)
```
kfwilson: do you still need df? You only have one measure_data now so I think you can just use that

tests/test_core.py
```diff
@@ -17,61 +17,69 @@ def test_get_ids_for_inconsistent_entities(cause_list, sequela_list):
 
 
 def test_get_ids_for_death(cause_list, sequela_list):
-    ids = core._get_ids_for_measure(cause_list, 'death')
+    ids = [core._get_ids_for_measure(c, 'death')[0] for c in cause_list]
     assert set(ids) == {c.gbd_id for c in cause_list}
     with pytest.raises(core.InvalidQueryError):
-        core._get_ids_for_measure(sequela_list, 'death')
+        for sequela in sequela_list:
```
kfwilson: I've never thought to do this in a loop before. Does pytest check that every iteration of the loop raises the error or just the first?

src/vivarium_inputs/core.py
michaella21: Hmm I'm not sure :( I honestly have not even seen the case that has a `sub_id`, thus not sure what makes those two differences. Since it's an existing code, maybe @collijk can answer?

src/vivarium_inputs/core.py
michaella21: yep, nice catch 👍

src/vivarium_inputs/core.py
michaella21: we can, let me add it to the error message. I just added this error so that the loader can catch the `InvalidQueryError`. Maybe it doesn't even need a error message. I wasn't sure who else other than us + artifact loader will call this function

src/vivarium_inputs/core.py
michaella21: right, so many lingering thins from the old code :(

tests/test_core.py
halfhorst: It stops after it catches the first! I just tested it

kfwilson: I guess that makes sense. @michaella21 maybe switch the pytest.raises and for then if you want to check that it fails on every sequela?

michaella21: We can do either. In general I put pytest -x to stop at the first failure since we want every test passed anyways. So I thought that it would be fine because this test will always break if there is any sequela in the list not raising the correct error, meaning it passes only if every sequela raises the error. But no strong preference. I can switch

kfwilson: I think it's just a matter of whether we think it's sufficient to test it on one sequela, in which case, we don't need the loop or if we want to test it on every sequela in the list, in which case, we should switch the pytest.raises and the loop because right now it's only checking the first in the list.

michaella21: I anyways switched, but just curious. @halfhorst Does pytest get out of the loop regardless it catches or not after checking the first sequela in the list? I thought that it loops through every sequela until it catches the first failure (or the end of the list). so, either pass all or check until it fails. If not, another shocking found from python :(

src/vivarium_inputs/core.py
```diff
@@ -83,15 +83,15 @@ class DuplicateDataError(DataError):
     pass
 
 
-def get_draws(entities: Sequence[ModelableEntity], measures: Iterable[str],
+def get_draws(entity: ModelableEntity, measures: Iterable[str],
```
collijk: Only take a single measure here.

src/vivarium_inputs/core.py
collijk: @kfwilson That's why it's and `InvalidQueryError`

collijk: I think the sub_id stuff is a hold over from a now extinct gbd_mapping format.  

A good rule of thumb if you don't think code corresponds to anything is to remove it and see if anything breaks downstream.

collijk: With the code in `vivarium_inputs` only~

src/vivarium_inputs/core.py
```diff
@@ -467,47 +456,37 @@ def _pull_rr_data_from_gbd(measure_ids):
...
-                    data = measure_data.copy()
-                    data['year_id'] = id
-                    missing_data = missing_data.append(data)
-                measure_data = pd.concat([measure_data] + missing_data)
                 measure_data['coverage_gap'] = entity.name
                 measure_data = measure_data.rename(columns={"rei_id": "risk_id"})
                 del measure_data['measure']
-                df.append(measure_data)
-        measure_data = pd.concat(df)
+            except NotADirectoryError:
```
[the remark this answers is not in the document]
michaella21: this is for coverage gap which doesn't have RR. `loader` calls all the measure for coverage_gap and if RR is not in the aux folder it gives a `NotADirectoryError` since the directory has just not been made yet.

src/vivarium_inputs/core.py
```diff
@@ -152,7 +152,7 @@ def get_draws(entities: Sequence[ModelableEntity], measures: Iterable[str],
 
 
 # TODO: Move to utilities.py
-def _get_ids_for_measure(entities: Sequence[ModelableEntity], measure: str) -> List:
+def _get_ids_for_measure(entity: ModelableEntity, measure: str) -> List:
```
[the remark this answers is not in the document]
michaella21: Hmm I fixed it as int at first then changed back to `[value]` since `vivarium_gbd_access.gbd` expects an iterable. Are we changing everything there also?

src/vivarium_inputs/core.py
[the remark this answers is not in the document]
collijk: Raise a warning here that says exactly that.   We don't want to silently return things the user doesn't expect.

src/vivarium_inputs/core.py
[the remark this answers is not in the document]
collijk: Yep!

collijk:
Superceded by PR#28

PR #23, opened 2018-10-22
handling apostrophes in loc names

kfwilson:
Allowing for apostrophes in loc names

Also cleaning up the file names for the artifacts so that they don't have spaces/apostrophes in them, e.g.,  an artifact for Cote d'Ivoire would be named as "config_name_Cote_d-Ivoire.hdf" instead of "config_name Cote d'Ivoire.hdf"

### DIFF
--- a/src/vivarium_inputs/data_artifact/cli.py
+++ b/src/vivarium_inputs/data_artifact/cli.py
@@ -35,9 +35,10 @@ def build_artifact(model_specification, locations, project,
     under the "proj_cost_effect" project unless a different project is
     specified. Multiple, optional LOCATIONS can be provided to overwrite
     the configuration file. For locations containing spaces, replace the
-    space with an underscore, e.g.:
+    space with an underscore and surround any locations containing 
+    apostrophes with double quotes, e.g.:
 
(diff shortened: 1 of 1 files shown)


src/vivarium_inputs/data_artifact/cli.py
```diff
@@ -65,11 +66,12 @@ def build_artifact(model_specification, locations, project,
     if num_locations > 0:
         script_args += "--location {}"
         for i, location in enumerate(locations):
+            location = location.replace("'", "-")
```
halfhorst: I think it might kinda look funny to have apostrophes replaced with hyphens. I feel like just removing them would be more readable / in keeping with how people normally abuse grammar

kfwilson: So I did the dash because I need to be able to later on put back the ' before it tries to build the artifact because that's how the names are in the location tables, but I can definitely change it in the filenames to just exclude entirely. The only reason I put it as a dash there was if there were ever the case where someone wanted to parse the artifact names to get the location names but I don't know how likely that is

PR #28, opened 2018-10-30
Feature/make args singular

collijk:
Top level items:

- Moved a bunch of utility functions to utilities.py
- Made most of the private functions not private anymore.  This is a library without a public interface right now.
- Made all the entity/measure/location arguments singular.
- General cleanup of a bunch of functions.

### DIFF
--- a/tests/test_core.py
+++ b/tests/test_core.py
@@ -153,7 +153,8 @@ def test_get_relative_risk(mocker):
 
 def test_get_population_attributable_fraction(mocker):
     gbd_mock = mocker.patch("vivarium_inputs.core.gbd")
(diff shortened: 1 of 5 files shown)


tests/test_core.py
```diff
@@ -225,7 +128,7 @@ def test_get_relative_risk(mocker):
     rr_ = pd.DataFrame(columns=draw_cols, index=pd.MultiIndex.from_product([*rr_maps.values()], names=[*rr_maps.keys()]))
     rr_[draw_cols] = np.random.random_sample((len(rr_), 10)) * 10
     gbd_mock.get_relative_risks.return_value = rr_.reset_index()
```
michaella21: actually the bug was here. since you changed the function all singular, it should be gbd_mock.get_relative_risk.return value on 130

PR #29, opened 2018-11-02
Refactor/paf by risk

michaella21:
Now we only pull PAF of continuous risk and etiology from core, thus removed all of custom special code for PAF. Also little changes on loader and tests to make things work.

### DIFF
--- a/src/vivarium_inputs/core.py
+++ b/src/vivarium_inputs/core.py
@@ -270,13 +270,13 @@ def get_exposure(entity, location_id):
     entity_id = get_id_for_measure(entity, 'exposure')
 
     if entity.kind in ['risk_factor', 'etiology']:
-        data = gbd.get_exposure(risk_id=entity_id, location_id=location_id)
+        data = gbd.get_exposure(entity_id=entity_id, location_id=location_id)
         measure_id = get_exposure_measure_id(data)
         if measure_id == get_measure_id('prevalence'):
             data = convert_exposure_prevalence_to_proportion(data)
         data = handle_exposure_from_gbd(data)
     elif entity in special_cases:
-        data = gbd.get_exposure(risk_id=entity_id, location_id=location_id)
+        data = gbd.get_exposure(entity_id=entity_id, location_id=location_id)
         data = handle_gbd_coverage_gap_data(entity, data, 0)
         data = handle_exposure_from_gbd(data)
         del data['modelable_entity_id']
--- a/src/vivarium_inputs/core.py
+++ b/src/vivarium_inputs/core.py
@@ -211,6 +211,8 @@ def get_relative_risk(entity: Union[Risk, CoverageGap], location_id: int):
 
     data = standardize_all_age_groups(data)
     data = standardize_data(data, 1)
+    if 318 in data['rei_id']:
+        data.loc[data['rei_id'] == 318, 'rei_id'] = 'low_measles_vaccine_coverage_first_dose'
     return data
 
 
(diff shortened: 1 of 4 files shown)


src/vivarium_inputs/core.py
```diff
@@ -234,23 +233,15 @@ def pull_rr_data_from_gbd(measure_id, location_id):
 
 
 def get_population_attributable_fraction(entity, location_id):
-    special_cases = {risk_factors.unsafe_water_source}
     entity_id = get_id_for_measure(entity, 'population_attributable_fraction')
 
-    if entity.kind == 'cause':
+    if entity.kind == 'etiology':
```
collijk: Both of these branches do the exact same thing.  Combine them?

tests/test_core.py
```diff
@@ -153,18 +154,30 @@ def test_get_relative_risk(mocker):
 
 
 def test_get_population_attributable_fraction(mocker):
+
+    categorical_risk = [r for r in risk_factors if r.distribution in ['polytomous','dichotomous']]
+    coverage_gap_list = [c for c in coverage_gaps]
+
+    for risk in categorical_risk:
```
collijk: Are you actually pulling all the data here?

src/vivarium_inputs/data_artifact/loaders.py
```diff
@@ -376,6 +373,24 @@ def get_risk_relative_risk(risk, location):
     return result
 
 
+def get_risk_population_attributable_fraction(risk, location):
+    if risk.distribution not in ['lognormal', 'normal', 'ensemble']:
+        result = None
```
collijk: This should raise an error, right? This function should only be called on continuous risks.  Move the check up into the main risk loader function.  Otherwise we'll end up storing keys with empty values in the artifact.

src/vivarium_inputs/data_artifact/loaders.py
```diff
@@ -376,6 +373,24 @@ def get_risk_relative_risk(risk, location):
...
+    else:
+        paf = core.get_draws(risk, 'population_attributable_fraction', location)
+        normalized = []
+        for key, group in paf.groupby(["cause_id"]):
```
collijk: Can we push any of this code back into core?  I don't think the loaders should be doing this kind of stuff.

src/vivarium_inputs/data_artifact/loaders.py
```diff
@@ -424,31 +439,17 @@ def get_coverage_gap_relative_risk(coverage_gap, location):
     if data.empty:
         data = None
     else:
-        data = handle_coverage_gap_rr_paf(data)
+        data = normalize(data)
```
collijk: This code too I think should be pushed back to core if possible.

tests/test_core.py
michaella21: it shouldn't pull the actual data. It's checking whether the function raises an error and not pull the data and none of them should pull the data since they are either categorical risk or coverage gaps.

michaella21:
It passed all the smoke tests with ceam_experiments branch which I think we will bring in after merging this first. (and also don't forget to pull the gbd_access)

src/vivarium_inputs/data_artifact/loaders.py
collijk: Still need to address this.

michaella21: huh yeah, I don't know why I put the error on line 135 instead of here. will change that

src/vivarium_inputs/data_artifact/loaders.py
```diff
@@ -336,8 +333,9 @@ def get_risk_relative_risk(risk, location):
 
 
 def get_risk_population_attributable_fraction(risk, location):
-    if risk.distribution not in ['lognormal', 'normal', 'ensemble']:
-        result = None
+    if risk.distribution not in ['normal', 'lognormal', 'ensemble']:
```
collijk: Oh, I made a mistake.  I think the error handling does  actually belong up in `get_risk_data`.  I was objecting to the setting results to `None` in this function.  I didn't realize it was unreachable code.

src/vivarium_inputs/core.py
```diff
@@ -212,6 +211,8 @@ def get_relative_risk(entity: Union[Risk, CoverageGap], location_id: int):
 
     data = standardize_all_age_groups(data)
     data = standardize_data(data, 1)
+    if 318 in data['rei_id']:
```
collijk: I didn't see this before.  

1. Why are we putting a name in an id column?
2. If we need to do this for something downstream, leave a comment about it.
3. Shouldn't this happen in the special_cases branch?

PR #30, opened 2018-11-10
give an option to build artifact locatlly within qlogin

michaella21:
It's for MIC-252. The current option to build artifacts for the list of locations now becomes `multi_build_artifact` and default `build_artifact` will run locally with a debugger option.

### DIFF
--- a/src/vivarium_inputs/data_artifact/cli.py
+++ b/src/vivarium_inputs/data_artifact/cli.py
@@ -16,11 +16,26 @@ from vivarium.config_tree import ConfigTree
 @click.command()
 @click.argument('model_specification', type=click.Path(dir_okay=False,
                 readable=True))
+@click.option('--output-root', '-o', type=click.Path(file_okay=False, writable=True),
+              help="Directory to save artifact to. "
+                   "Overwrites model specification file")
+@click.option('--append', '-a', is_flag=True,
+              help="Preserve existing artifact and append to it")
+@click.option('--verbose', '-v', is_flag=True,
+              help="Turn on debug mode for logging")
+@click.option('--pdb', 'debugger', is_flag=True, help='Drop the debugger if an error occurs')
+def build_artifact(model_specification, output_root, append, verbose, debugger):
+    _build_artifact()
+
+
+@click.command()
+@click.argument('model_specification', type=click.Path(dir_okay=False,
+                readable=True))
 @click.argument('locations', nargs=-1)
 @click.option('--project', '-P', default='proj_cost_effect',
               help='Cluster project under which the job will '
                    'be submitted. Defaults to proj_cost_effect')
-@click.option('--output_root', '-o', type=click.Path(file_okay=False, writable=True),
+@click.option('--output-root', '-o', type=click.Path(file_okay=False, writable=True),
               help="Directory to save artifact to. "
                    "Overwrites model specification file")
 @click.option('--append', '-a', is_flag=True,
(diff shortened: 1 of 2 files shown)


src/vivarium_inputs/data_artifact/cli.py
```diff
@@ -13,14 +13,29 @@
...
+@click.option('--output-root', '-o', type=click.Path(file_okay=False, writable=True),
+              help="Directory to save artifact to. "
+                   "Overwrites model specification file")
+@click.option('--append', '-a', is_flag=True,
+              help="Preserve existing artifact and append to it")
+@click.option('--verbose', '-v', is_flag=True,
+              help="Turn on debug mode for logging")
+@click.option('--pdb', 'debugger', is_flag=True, help='Drop the debugger if an error occurs')
+def build_artifact(model_specification, output_root, append, verbose, debugger):
+    _build_artifact()
```
collijk: Why are you wrapping `_build_artifact` instead of replacing it with this function?

src/vivarium_inputs/data_artifact/cli.py
```diff
@@ -13,14 +13,29 @@
...
+
 @click.command()
 @click.argument('model_specification', type=click.Path(dir_okay=False,
                 readable=True))
 @click.argument('locations', nargs=-1)
 @click.option('--project', '-P', default='proj_cost_effect',
               help='Cluster project under which the job will '
                    'be submitted. Defaults to proj_cost_effect')
-@click.option('--output_root', '-o', type=click.Path(file_okay=False, writable=True),
+@click.option('--output-root', '-o', type=click.Path(file_okay=False, writable=True),
```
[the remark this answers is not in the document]
michaella21: This one works. I'm sure Kate tested before since she asked me to change this while I was working on this. I just tested also. it shouldn't break any

src/vivarium_inputs/data_artifact/cli.py
michaella21: Initially it was because I didn't think too hard. Then i tried to replace it today, which actually causes some problem. Instead of wrapping it, I just add click arguments on the `_build_artifact` and replace it as `build_artifact`. Then it worked fine for the local build (i.e., build for one location within qlogin) but when I tried to do multi build with multiple locations, click gets encoding error like, 

``` 
Traceback (most recent call last):
  File "/ihme/homes/hpark3/simulation/vivarium_inputs/src/vivarium_inputs/data_artifact/cli.py", line 371, in <module>
    build_artifact()
  File "/ihme/code/hpark3/miniconda3/envs/vivarium/lib/python3.6/site-packages/click/core.py", line 764, in __call__
    return self.main(*args, **kwargs)
  File "/ihme/code/hpark3/miniconda3/envs/vivarium/lib/python3.6/site-packages/click/core.py", line 696, in main
    _verify_python3_env()
  File "/ihme/code/hpark3/miniconda3/envs/vivarium/lib/python3.6/site-packages/click/_unicodefun.py", line 124, in _verify_python3_env
    ' mitigation steps.' + extra
RuntimeError: Click will abort further execution because Python 3 was configured to use ASCII as encoding for the environment. Consult https://click.palletsprojects.com/en/7.x/python3/ for mitigation steps.
...

Click discovered that you exported a UTF-8 locale
but the locale system could not pick up from it because
it does not exist.  The exported locale is "en_US.UTF-8" but it
is not supported
```
This didn't happen when the original code uses argparse  but it happens when argument is passed in through click.

michaella21: NVM, we can just use the existing function. I don't know why I thought things in more complicated way

PR #34, opened 2018-11-28
Feature/forecasting data

kfwilson:
Preliminary code to start formatting forecasting data for sim. Only currently works for population, deaths, asfr, and cause specific mortality rates - we would need to get specific versions from forecasting for all other measures. 

Ran test simulation with age specific fertility, mortality, population, and csmr for peripheral artery disease out to 2040 using forecast data. 

I patched the forecasting data into an artifact. We need to think about when we want to use forecasting data/how to signal to the data artifact to pull forecasting data instead of gbd data.


src/vivarium_inputs/data_artifact/utilities.py
```diff
@@ -17,6 +17,6 @@ def normalize(data: pd.DataFrame) -> pd.DataFrame:
 
     estimation_years = core.get_estimation_years()
     year_start, year_end = min(estimation_years), max(estimation_years)
-    if "year" in data:
```
collijk: This change seems unrelated to forecasting data.  Were things broken before?

PR #37, opened 2018-12-18
Feature/cause level disability weights

halfhorst:


src/vivarium_inputs/core.py
```diff
@@ -171,17 +173,41 @@ def get_excess_mortality(entity, location_id):
...
+        sequelae = entity.sequelae
+
+        aggregate_dw = []
+        for seq in sequelae:
+            seq_prevalence = get_prevalence(seq, location_id).reset_index(drop=True)
+            seq_disability = disability_weights.loc[seq.healthstate.gbd_id == disability_weights.healthstate_id, :]
+            seq_disability = pd.concat([seq_disability[draw_columns]] * seq_prevalence.shape[0], ignore_index=True)
+            seq_prevalence[draw_columns] = seq_prevalence[draw_columns].mul(seq_disability, axis='columns')
+            seq_prevalence = seq_prevalence[id_columns + draw_columns]
+            melted_seq_prevalence = seq_prevalence.melt(id_vars=id_columns, var_name='draw')
```
collijk: To do with better use of indices and `sum`

src/vivarium_inputs/core.py
```diff
@@ -171,19 +174,54 @@ def get_excess_mortality(entity, location_id):
...
     else:
         raise DataMissingError(f"No disability weight available for the sequela {entity.name}")
+    data.loc[:, 'sequela_id'] = entity.gbd_id
+
+    return data
+
+
+def get_disability_weight(entity: Union[Sequela, Cause], location_id: int = None):
+    if entity.kind == 'sequela':
+        data = _get_sequela_disability_weights(entity, location_id)
```
kfwilson: it looks like _get_sequela_disability_weights only takes a single argument of the entity

src/vivarium_inputs/core.py
```diff
@@ -171,19 +174,58 @@ def get_excess_mortality(entity, location_id):
...
+
+    elif entity.kind == "cause":
+        if location_id is None:
```
kfwilson: Is this check necessary? None of the other core functions that need location have it and everything should (I think?) go through get_draws which will break if location is none

src/vivarium_inputs/core.py
```diff
@@ -171,19 +174,58 @@ def get_excess_mortality(entity, location_id):
...
-    return data.reset_index(drop=True)
+            # duplicate disability weight to make multiplication easier.
+            seq_disability = pd.concat([seq_disability[draw_columns]] * seq_prevalence.shape[0], ignore_index=True)
+
+            seq_prevalence.loc[:, draw_columns] = seq_prevalence[draw_columns] * seq_disability
+            seq_prevalence = seq_prevalence.set_index(id_columns)
+            sequela_level_data.append(seq_prevalence)
+
+        if len(sequela_level_data) == 0:
+            raise DataMissingError(f"The cause {entity.name} does not have enough sequela-levle information to generate "
```
kfwilson: Maybe clarify this message to say that the cause doesn't have any sequela-level info - if I got this error message, I would think that there were say 2 sequela but 3 were required or something like that

halfhorst: that's fair

src/vivarium_inputs/core.py
halfhorst: I just threw it in because this fxn can be called with and without the location_id so I wanted to validate, but I can let it fall through to get_draws and fail as well

michaella21: maybe need to add an error somewhere else, in the current setting, it will fail when the get_draws calls the get_location_id, which probably is a keyerror with not provided key for dictionary

src/vivarium_inputs/core.py
```diff
@@ -171,19 +174,56 @@ def get_excess_mortality(entity, location_id):
...
-    return data.reset_index(drop=True)
+            # duplicate disability weight to make multiplication easier.
+            seq_disability = pd.concat([seq_disability[draw_columns]] * seq_prevalence.shape[0], ignore_index=True)
+
+            seq_prevalence.loc[:, draw_columns] = seq_prevalence[draw_columns] * seq_disability
+            seq_prevalence = seq_prevalence.set_index(id_columns)
+            sequela_level_data.append(seq_prevalence)
+
+        if len(sequela_level_data) == 0:
+            raise DataMissingError(f"The cause {entity.name} does not have both a disability weight and prevalence for"
```
kfwilson: Wait I think this is even more confusing now. Because if any single sequela doesn't have disability weight, it just skips it (that's your comment saying take disability weight to be zero) right? and if any single sequela doesn't have prevalence, it will (I think?) break because there's no try/catch around get_prevalence. So if sequela_level_data is indeed length 0, it can only be because either the cause has no sequelae OR there is no disability weight for ALL sequelae of the cause

michaella21: yeah I think it will break since seq_disability is not defined in that case

src/vivarium_inputs/core.py
```diff
@@ -171,19 +174,52 @@ def get_excess_mortality(entity, location_id):
...
 
-    return data.reset_index(drop=True)
+            # duplicate disability weight to make multiplication easier.
+            seq_disability = pd.concat([seq_disability[draw_columns]] * seq_prevalence.shape[0], ignore_index=True)
+
+            seq_prevalence.loc[:, draw_columns] = seq_prevalence[draw_columns] * seq_disability
+            seq_prevalence = seq_prevalence.set_index(id_columns)
+            sequela_level_data.append(seq_prevalence)
+
+        data = reduce(lambda x, y: x + y, sequela_level_data)
```
collijk: `data = sum(sequela_level_data)`?

PR #40, opened 2018-12-24
Feature/aggregate multil loc artifacts

michaella21:
Add the aggregation for `multi_build_artifact`. It just combines the current multiple artifacts into one single artifact for the artifact which has the same set of keyspace. To support the appending option, also added the disaggregation process which is called whenever append option is selected. It should be able to handle all cases that a user wants (1) to add more locations, (2) to add more model specifications, (3) both (1) and (2). Tests for the logic of aggregation/disaggregation are needed, which will come also.


src/vivarium_inputs/data_artifact/aggregation.py
```diff
@@ -0,0 +1,127 @@
...
+    for a in artifacts:
+        if set(a.load('metadata.keyspace')) == keyspace_set:
+            valid_artifacts.append(a)
+            valid_locations.extend(a.load('metadata.locations'))
+
+    if set(valid_locations) < set(locations):
+        warnings.warn(f'Individual artifacts failed for {set(locations).difference(set(valid_locations))} '
+                      f'and only rest of locations will be aggregated')
+
+    metadata = {'keyspace': 'metadata.keyspace', 'location': 'metadata.locations', 'versions': 'metadata.versions'}
```
kfwilson: why do you need this dictionary? It seems like you only ever use the values and not the keys

src/vivarium_inputs/data_artifact/aggregation.py
```diff
@@ -0,0 +1,127 @@
...
+
+        for loc in existing_locations:
+            temp_path = f'{initial_artifact_path.parent.as_posix()}/{config_name}_{loc.replace(" ", "_")}.hdf'
+
+            hdf.touch(temp_path, False)
+            hdf.write(temp_path, EntityKey("metadata.versions"), current_versions)
+            hdf.write(temp_path, EntityKey("metadata.locations"), [loc])
+            hdf.write(temp_path, EntityKey('metadata.keyspace'), [EntityKey(k) for k in metadata.values()])
+            
+            new_artifact = Artifact(temp_path)
```
kfwilson: add the single location as a filter term to the artifact when you make it and then just load the data - the filter terms get used to filter before data is actually read in so it's probably faster esp. for artifacts with lots of locs to filter that way rather than reading everything in first and then filtering to loc

src/vivarium_inputs/data_artifact/builder.py
```diff
@@ -36,6 +40,14 @@ def setup(self, builder):
 
         builder.event.register_listener('post_setup', self.end_processing)
 
+    def write_metadata(self, path):
+        hdf.write(path, EntityKey("metadata.versions"),
+                  {k: pkg_resources.get_distribution(k).version for k in
+                   ['vivarium', 'vivarium_public_health', 'gbd_mapping', 'vivarium_inputs']})
+        hdf.write(path, EntityKey("metadata.locations"), [self.location])
+        hdf.write(path, EntityKey('metadata.keyspace'),
+                  [EntityKey('metadata.keyspace'), EntityKey('metadata.locations')])
```
kfwilson: I thought we had talked about having the keyspace be a list of string keys - is there a reason you decided to change to a list of entity keys?

src/vivarium_inputs/data_artifact/cli.py
```diff
@@ -61,8 +62,7 @@ def build_artifact(model_specification, output_root, append, verbose, debugger):
               help="Write SGE error logs to output location")
 @click.option('--memory', '-m', default=10, help="Specifies the amount of memory in G that will be requested for a "
                                                  "job. Defaults to 10. Only applies to the new cluster.")
-def multi_build_artifact(model_specification, locations, project,
-                   output_root, append, verbose, error_logs, memory):
+def multi_build_artifact(model_specification, locations, project, output_root, append, verbose, error_logs, memory):
```
kfwilson: update the doc string to clarify that the result is a single artifact containing multiple locations

src/vivarium_inputs/data_artifact/builder.py
michaella21: nope, this line was just written at the very beginning and then I didn't change it back to the strings. Then it was always overridden as strings during building artifacts so I didn't notice that it was like this :( well, I even forgot to add metadata.versions there :(

src/vivarium_inputs/data_artifact/cli.py
```diff
@@ -91,32 +91,54 @@ def multi_build_artifact(model_specification, locations, project,
...
+            for i, location in enumerate(locations):
+                location = location.replace("'", "-")
+                job_name = f"{config_path.stem}_{location}_build_artifact"
+                
+                command = build_submit_command(python_context_path, job_name, project, error_log_dir,
+                                               script_arguments.format(location), memory, archive=True, queue='all.q')
+                click.echo(f"submitting job {i+1} of {num_locations} ({job_name})")
+                jobid = submit_job(command, job_name)
+                jids.append(jobid)
+        else:
```
kfwilson: does this actually need to be a separate block? It seems like the only difference is the click message and whether or not the file name has the loc in it but that file is only temp anyways.

src/vivarium_inputs/data_artifact/aggregation.py
```diff
@@ -0,0 +1,127 @@
...
+    config_path = Path(args.config_path)
+    locations = args.locations
+    locations = [k.replace('_', ' ') for k in locations]
+    output = args.output_root
+    individual_artifacts = Path(output).glob('*.hdf')
+    artifacts = [Artifact(a.as_posix()) for a in individual_artifacts]
+
+    keyspace_set = set().union(*[a.load('metadata.keyspace') for a in artifacts])
+
+    valid_artifacts, valid_locations = [], []
```
kfwilson: I think technically someone could end up here with just a single argument (e.g., if they called multi_build_artifact but only gave a single location) so it might be worth checking if valid_artifacts is longer than 1 and only copying key by key if so. Otherwise, just rename the single artifact file which will be a lot quicker

src/vivarium_inputs/data_artifact/cli.py
michaella21: I didn't think it too hard and just wrap up the existing code as a function. But it seems that we should have removed it when we(actually I, so my fault) added the build_artifact option. Since now we don't have a case of multi_build_artifact with no location given. I will change that

PR #41, opened 2018-12-27
Refactor/change in hdf creation

michaella21:
This is another bit of current VPH PR#51 before bringing in the multiple location artifacts. With this change, appending logic only stays in the builder and artifact will create hdf only if the given file path is not a file

### DIFF
--- a/src/vivarium_inputs/data_artifact/builder.py
+++ b/src/vivarium_inputs/data_artifact/builder.py
@@ -59,26 +59,27 @@ class ArtifactBuilder:
 
             try:
                 artifact = Artifact(path, filter_terms=[f'draw == {draw}', get_location_term(location)])
-                if EntityKey('metadata.locations') not in artifact:
-                    warnings.warn('We will build from scratch', OutdatedArtifactWarning)
-                    artifact = create_new_artifact(path, draw, location)
-
-                elif artifact.load('metadata.locations') != [location]:
-                    raise ValueError(f"Artifact has {artifact.load('metadata.locations')} and we cannot append {location}")
-
-                if EntityKey('metadata.versions') not in artifact:
-                    warnings.warn('We will build from scratch', OutdatedArtifactWarning)
-                    artifact = create_new_artifact(path, draw, location)
-
-                elif artifact.load('metadata.versions') != get_versions():
-                    warnings.warn('Your artifact was made under different versions. We will wipe it out', DeprecationWarning)
-                    artifact = create_new_artifact(path, draw, location)
 
             except NoSuchNodeError:
                 #  it means that path was a file but does not have metadata.keyspace inside
                 warnings.warn('We will wipe it out and build from scratch', OutdatedArtifactWarning)
                 artifact = create_new_artifact(path, draw, location)
 
+            if EntityKey('metadata.locations') not in artifact:
+                warnings.warn('We will build from scratch', OutdatedArtifactWarning)
+                artifact = create_new_artifact(path, draw, location)
+
+            elif artifact.load('metadata.locations') != [location]:
+                raise ValueError(f"Artifact has {artifact.load('metadata.locations')} and we cannot append {location}")
+
+            if EntityKey('metadata.versions') not in artifact:
+                warnings.warn('We will build from scratch', OutdatedArtifactWarning)
+                artifact = create_new_artifact(path, draw, location)
+
+            elif artifact.load('metadata.versions') != get_versions():
+                warnings.warn('Your artifact was made under different versions. We will wipe it out',
+                              OutdatedArtifactWarning)
+                artifact = create_new_artifact(path, draw, location)
         else:
             artifact = create_new_artifact(path, draw, location)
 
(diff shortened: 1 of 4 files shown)


src/vivarium_inputs/data_artifact/builder.py
```diff
@@ -23,11 +26,16 @@ class ArtifactBuilder:
     def setup(self, builder):
         path = builder.configuration.input_data.artifact_path
         append = builder.configuration.input_data.append_to_artifact
-        hdf.touch(path, append)
+
         draw = builder.configuration.input_data.input_draw_number
 
         self.location = builder.configuration.input_data.location
+        if not append:
```
kfwilson: Is this block necessary? You've changed Artifact now so that if you give it a path that isn't already a file, it creates a new artifact. Do we want to do the overwriting of existing artifacts in the builder? That seems a little bit of a weird place to me

michaella21: because... now there's no way to tell the difference whether you want to wipe out and recreate the file vs you want to append it. Now appending logic only stays here and Artifact side doesn't know whether you want to keep the file or overwrite the file. Artifact will only make a new hdf if the given file path is not a file. So if you want to delete the current one and just overwrite it, it needs to explicitly make a new file.

kfwilson: Yeah that makes sense to keep the append only in one place

src/vivarium_inputs/data_artifact/builder.py
```diff
@@ -36,6 +44,22 @@ def setup(self, builder):
 
         builder.event.register_listener('post_setup', self.end_processing)
 
+    def write_metadata(self, append):
+        if append:
+            try:
```
collijk: As a general rule of thumb, always try to push preconditions up as early as possible and don't hide them in function calls that don't clearly indicate that you're checking preconditions.  I'd suggest you make a function like 
```
def initialize_artifact(path, location, draw, append):
    if append:
        a = Artifact(path, filter_terms=[f'draw == {draw}', get_location_term(self.location)])
        if 'metadata.location' not in a:
            log that it's an old artifact
            unlink path
            a = Artifact(...)
        elif a.load('metadata.location') != location:
            raise error

        if 'metadata.versions' not in a:
            log that it's an old artifact
            path.unlink()
            a = Artifact(...)
        elif a.load('metadata.versions') != get_versions():
             either wipe artifact or raise error or warn?
    else:
        a = Artifact(...)
        a.write(versions)
        a.write(location)

    return a
```

collijk: probably as a static method

collijk: This will give us a single place to put deprecations and think about the artifact metadata.

src/vivarium_inputs/data_artifact/builder.py
```diff
@@ -36,6 +39,43 @@ def setup(self, builder):
 
         builder.event.register_listener('post_setup', self.end_processing)
 
+    @staticmethod
+    def initialize_artifact(path, append, draw, location):
+        def create_new_artifact():
```
collijk: The `initialize_artifact` method will probably be much easier to test if you pull out the `create_new_artifact` function.

src/vivarium_inputs/data_artifact/builder.py
```diff
@@ -1,18 +1,27 @@
...
 
 from vivarium_inputs.data_artifact.loaders import loader
 from vivarium_inputs.forecasting import load_forecast
+from vivarium_inputs.data_artifact.utilities import get_versions
 
 
 _log = logging.getLogger(__name__)
 
 
+class OutdatedArtifactWarning(Warning):
```
collijk: You made this class but aren't using it.

src/vivarium_inputs/data_artifact/builder.py
```diff
@@ -36,6 +44,46 @@ def setup(self, builder):
...
+
+                if EntityKey('metadata.versions') not in artifact:
+                    warnings.warn('We will build from scratch', OutdatedArtifactWarning)
+                    artifact = create_new_artifact(path, draw, location)
+
+                elif artifact.load('metadata.versions') != get_versions():
+                    warnings.warn('Your artifact was made under different versions. We will wipe it out', DeprecationWarning)
+                    artifact = create_new_artifact(path, draw, location)
+
+            except NoSuchNodeError:
```
collijk: Why would you have an if sieve for locations and versions but treat the keyspace differently in a try/except block?

src/vivarium_inputs/data_artifact/builder.py
collijk: Or, rather, not using it consistently.  You still have a deprecation warning hanging around

src/vivarium_inputs/data_artifact/builder.py
michaella21: because if we can't initialize the Artifact itself without keyspace. So I can't just create the object and check whether it has a keyspace or not

PR #42, opened 2019-01-01
Feature/aggregate multi loc artifacts

michaella21:
Re-open the PR for aggregating artifacts. Pretty sure that there would be a more efficient way to do this. For now, mechanically working. Tested all possible cases that I could think of. But please review them carefully in case that I missed any edge case. Travis will fail but tests should work locally. (I thought that it was a problem with permission/tmpdir and wasted 1+ day to fork around the travis. But i concluded that it wasn't)


src/vivarium_inputs/data_artifact/aggregation.py
```diff
@@ -0,0 +1,166 @@
...
+            else:
+                assert np.all([d.equals(data[0]) for d in data])
+                data = data[0]
+            artifact.write(k, data)
+        else:
+            assert np.all([d == data[0] for d in data])
+            artifact.write(k, data[0])
+
+    # clean-up
+    for loc in valid_locations:
```
kfwilson: valid_locations only includes the artifacts that succeeded - clean up failed ones too

src/vivarium_inputs/data_artifact/aggregation.py
```diff
@@ -0,0 +1,166 @@
...
+    return artifact
+
+
+def main(rawargs=None):
+    """ Aggregate multiple artifacts to a single artifact.
+        We only take the union of the each keyspace of single artifacts and
+        do not aggregate any single artifact does not have all the keys
+        in the union of keyspaces.
+
+        Aggregation is held until single artifact building jobs are completed.
```
kfwilson: this holding doesn't happen here so it's a little confusing to put it in the docstring

src/vivarium_inputs/data_artifact/aggregation.py
```diff
@@ -0,0 +1,166 @@
...
+    """
+    parser = argparse.ArgumentParser()
+    config_path, locations, output = unpack_arguments(parser, rawargs)
+
+    individual_artifacts = []
+    for loc in locations:
+        individual_path = Path(output)/f'{config_path.stem}_{loc.replace(" ", "_")}.hdf'
+        individual_artifacts.append(individual_path)
+    
+    artifacts = [Artifact(a.as_posix()) for a in individual_artifacts]
```
kfwilson: maybe do this as part of the previous loop so you don't have to loop over loc list twice

src/vivarium_inputs/data_artifact/aggregation.py
```diff
@@ -0,0 +1,166 @@
...
+                      ArtifactAggregationWarning)
+
+        initial_artifact_path.unlink()
+        existing_locations = {}
+    else:
+        existing_keys = existing_artifact.load('metadata.keyspace')
+        existing_locations = existing_artifact.load('metadata.locations')
+
+        for loc in existing_locations:
+            location = get_location_term(loc)
```
kfwilson: get_location_term also includes 'Global' in the filter so I think if you have any data that has a location of 'Global', you'll end up with as many copies of it as you have locations, which will cause duplicate errors if that data is read in and used in any interpolations

src/vivarium_inputs/data_artifact/aggregation.py
```diff
@@ -0,0 +1,166 @@
...
+
+    if not initial_artifact_path.is_file():
+        raise ArtifactException(
+            f'To append it, you should provide the existing artifact. {output_root}/{config_name}.hdf does not exist')
+
+    existing_artifact = Artifact(initial_artifact_path.as_posix())
+
+    if existing_artifact.load('metadata.versions') != get_versions():
+        #  FIXME: For now we only warn and build from scratch. We need a smarter way to handle ths.
+        warnings.warn('Your artifact was built under the different versions and cannot append it',
```
kfwilson: I think it would be helpful to explicitly say in this warning 'a new artifact will be built' - I think if I saw this warning, I might be confused whether the build was still happening or not

src/vivarium_inputs/data_artifact/aggregation.py
```diff
@@ -82,6 +91,16 @@ def aggregate(artifacts: [Artifact], artifact_path: str) -> Artifact:
     for loc in valid_locations:
         f = artifact_path.parent/f'{artifact_path.stem}_{loc.replace(" ", "_")}.hdf'
         f.unlink()
+
+    invalid_locations = set(locations).difference(set(valid_locations))
+    if len(invalid_locations) > 0:
+        new_dir = artifact_path.parent/'broken_artifacts'
+        new_dir.mkdir()
+
+        for loc in invliad_locations:
```
kfwilson: typo here - should be invalid_locations

src/vivarium_inputs/data_artifact/cli.py
```diff
@@ -97,8 +97,15 @@ def multi_build_artifact(model_specification, locations, project, output_root, a
     config_path = pathlib.Path(model_specification).resolve()
     python_context_path = pathlib.Path(shutil.which("python")).resolve()
     script_path = pathlib.Path(__file__).resolve()
+    locations = [l.replace("'", "-") for l in locations]
```
kfwilson: I think you want to do this replace after you compare with existing locations because the existing locations will still have apostrophes

PR #44, opened 2019-01-04
Feature/gbd 2017 update sequela

michaella21:
Just keep it open in case if someone wants to make a change/to edit/to remove or to add new things in extract.py. There are obvious repeated codes which I believe that we will clean up later once we have more things coming in. Commenting out in init is necessary to test any new code and to avoid an import error.

### DIFF
--- a/src/vivarium_inputs/extract.py
+++ b/src/vivarium_inputs/extract.py
@@ -72,8 +72,8 @@ def get_sequela_disability_weight(entity: Sequela, _) -> pd.DataFrame:
 def check_years(df: pd.DataFrame, year_type: str):
     years = {'annual': [y for y in range(1990, 2018)], 'binned': gbd.get_estimation_years()}
     expected_years = years[year_type]
-    if set(df.year_id.unique) < expected_years:
-        raise DataNotExistError(f'Data has missing years: {set(expected_years).difference(set(df.year_id.unique))}')
+    if set(df.year_id.unique()) < expected_years:
+        raise DataNotExistError(f'Data has missing years: {set(expected_years).difference(set(df.year_id.unique()))}')
 
 
 def check_columns(expected_cols:List, existing_cols:List):
--- a/src/vivarium_inputs/extract.py
+++ b/src/vivarium_inputs/extract.py
@@ -1,0 +1,83 @@
+import pandas as pd
+import warnings
+from typing import List
+
+from vivarium_gbd_access import gbd
+from gbd_mapping.sequela import Sequela
+
+from .utilities import DataNotExistError, DataAbnormalError
+
+
+DRAW_COLS = [f'draw_{i}' for i in range(1000)]
+COLS = {
+    'sequela':{
+        'prevalence':
+            ['year_id', 'age_group_id', 'sex_id', 'measure_id', 'sequela_id', 'location_id', 'metric_id'] + DRAW_COLS,
+        'incidence':
+            ['year_id', 'age_group_id', 'sex_id', 'measure_id', 'sequela_id', 'location_id', 'metric_id'] + DRAW_COLS,
+        'disability_weight':
+            ['location_id', 'age_group_id', 'sex_id', 'measure', 'healthstate_id', 'healthstate'] + DRAW_COLS
+    }
+}
+
+
+def get_sequela_prevalence(entity: Sequela, location_id: int) -> pd.DataFrame:
+    if not entity.prevalence_exists:
+        raise DataNotExistError(f'{entity.name} does not have prevalence data')
+    if not entity.prevalence_in_range:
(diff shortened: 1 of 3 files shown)


src/vivarium_inputs/extract.py
```diff
@@ -0,0 +1,76 @@
...
+
+def get_sequela_prevalence(entity: Sequela, location_id: int) -> pd.DataFrame:
+    if not entity.prevalence_exists:
+        raise DataNotExistError(f'{entity.name} does not have prevalence data')
+    if not entity.prevalence_in_range:
+        raise warnings.warn(f'{entity.name} has prevalence but its range is abnormal')
+    data = gbd.get_como_draws(entity_id=entity.gbd_id, location_id=location_id, entity_type='sequela')
+    data = data[data.measure_id == 5]
+
+    assert data.metric_id.unique() == [3], 'prevalence should have only rate (metric_id 3)'
```
kfwilson: maybe instead of asserting, raise a DataAbnormalError here?

src/vivarium_inputs/extract.py
```diff
@@ -0,0 +1,76 @@
...
+    expected_cols = COLS['sequela']['disability_weight']
+    check_columns(expected_cols, data.columns)
+    return data
+
+
+def check_years(df: pd.DataFrame, year_type: str):
+    years = {'annual': [y for y in range(1990, 2018)], 'binned': gbd.get_estimation_years()}
+    expected_years = years[year_type]
+    if set(df.year_id.unique()) < set(expected_years):
+        raise DataNotExistError(f'Data has missing years: {set(expected_years).difference(set(df.year_id.unique()))}')
```
kfwilson: I would think this should be a DataAbnormalError because the data does exist, it just has weird years

src/vivarium_inputs/extract.py
```diff
@@ -0,0 +1,76 @@
...
+
+def get_sequela_incidence(entity: Sequela, location_id: int) -> pd.DataFrame:
+    if not entity.incidence_exists:
+        raise DataNotExistError(f'{entity.name} does not have incidence data')
+    if not entity.incidence_in_range:
+        raise warnings.warn(f'{entity.name} has incidence but its range is abnormal')
+    data = gbd.get_como_draws(entity_id=entity.gbd_id, location_id=location_id, entity_type='sequela')
+    data = data[data.measure_id == 6]
+
+    assert data.metric_id.unique() == [3], 'incidence should have only rate (metric_id 3)'
```
kfwilson: same comment as above about this assert

src/vivarium_inputs/extract.py
```diff
@@ -0,0 +1,76 @@
...
+from gbd_mapping.sequela import Sequela
+
+from .utilities import DataNotExistError, DataAbnormalError
+
+
+DRAW_COLS = [f'draw_{i}' for i in range(1000)]
+COLS = {
+    'sequela':{
+        'prevalence':
+            ['year_id', 'age_group_id', 'sex_id', 'measure_id', 'sequela_id', 'location_id', 'metric_id'] + DRAW_COLS,
```
vivarium-dev: Make demography cols

src/vivarium_inputs/extract.py
```diff
@@ -0,0 +1,76 @@
+import pandas as pd
+import warnings
+from typing import List
+
+from vivarium_gbd_access import gbd
+from gbd_mapping.sequela import Sequela
+
+from .utilities import DataNotExistError, DataAbnormalError
+
```
vivarium-dev: Make a dict for measure and metric ids.  use db_queries.get_ids('measure'/'metric')

src/vivarium_inputs/extract.py
```diff
@@ -0,0 +1,76 @@
...
+        raise DataNotExistError(f'{entity.name} does not have data for disability weight')
+    disability_weights = gbd.get_auxiliary_data('disability_weight', 'sequela', 'all')
+    data = disability_weights.loc[disability_weights.healthstate_id == entity.healthstate.gbd_id, :]
+    expected_cols = COLS['sequela']['disability_weight']
+    check_columns(expected_cols, data.columns)
+    return data
+
+
+def check_years(df: pd.DataFrame, year_type: str):
+    years = {'annual': [y for y in range(1990, 2018)], 'binned': gbd.get_estimation_years()}
```
collijk: Look about getting all years from gbd tooling

src/vivarium_inputs/utilities.py
```diff
@@ -0,0 +1,13 @@
+
```
collijk: Start with VivariumInputsException

class DataNotExistError(VivariumInputsException, FileNotFoundError):
    pass

class DataAbnormalError(VivariumInputsException, ValueError):
    pass

PR #46, opened 2019-01-04
Fix bug in aggregation test

collijk:


collijk:
It actually does pass with the correct branches.

PR #62, opened 2019-01-07
Add covariate functions

collijk:


src/vivarium_inputs/extract.py
```diff
@@ -91,7 +91,8 @@ def extract_mediation_factors(entity, location_id: int) -> pd.DataFrame:
 
 
 def extract_estimate(entity, location_id: int) -> pd.DataFrame:
-    raise NotImplementedError()
+    data = gbd.get_covariate_estimate(entity.gbd_id, location_id)
```
michaella21: You need to add `check_metadata` since not every covariate have all three columns and we have flags

src/vivarium_inputs/validation/raw.py
```diff
@@ -149,7 +149,11 @@ def _validate_mediation_factors(data, entity, location_id):
 
 
 def _validate_estimate(data, entity, location_id):
-    raise NotImplementedError()
+    expected_columns = ['model_version_id', 'covariate_id', 'covariate_name_short', 'location_id',
+                        'location_name', 'year_id', 'age_group_id', 'age_group_name', 'sex_id',
+                        'sex', 'mean_value', 'lower_value', 'upper_value']
```
michaella21: does every covariate contain age_group_id and age_group_name?

PR #71, opened 2019-01-16
Feature/gbd 2017 check metadata

kfwilson:
This updates the metadata checks based on all the mapping updates - throws warnings for pretty much everything because of potential location variation. It also includes a few stray cleanups to error/warning messages to include punctuation :)

### DIFF
--- a/src/vivarium_inputs/validation/raw.py
+++ b/src/vivarium_inputs/validation/raw.py
@@ -240,7 +240,7 @@ def _validate_exposure_distribution_weights(data, entity, location_id):
 
 
 def _validate_relative_risk(data, entity, location_id):
-    del entity
+    del entity  # unused
     expected_columns = ('rei_id', 'modelable_entity_id', 'cause_id', 'mortality',
                         'morbidity', 'metric_id', 'parameter') + DEMOGRAPHIC_COLUMNS + DRAW_COLUMNS
     check_columns(expected_columns, data.columns)
@@ -249,7 +249,7 @@ def _validate_relative_risk(data, entity, location_id):
 
 
 def _validate_population_attributable_fraction(data, entity, location_id):
-    del entity
+    del entity  # unused
     expected_columns = ('metric_id', 'measure_id', 'rei_id', 'cause_id') + DRAW_COLUMNS + DEMOGRAPHIC_COLUMNS
     check_columns(expected_columns, data.columns)
     check_years(data, 'annual')
@@ -261,7 +261,7 @@ def _validate_mediation_factors(data, entity, location_id):
 
 
 def _validate_estimate(data, entity, location_id):
-    del entity
+    del entity  # unused
     expected_columns = ['model_version_id', 'covariate_id', 'covariate_name_short', 'location_id',
                         'location_name', 'year_id', 'age_group_id', 'age_group_name', 'sex_id',
                         'sex', 'mean_value', 'lower_value', 'upper_value']
@@ -279,7 +279,7 @@ def _validate_utilization(data, entity, location_id):
 
 
 def _validate_structure(data, entity, location_id):
-    del entity
+    del entity  # unused
     expected_columns = ['age_group_id', 'location_id', 'year_id', 'sex_id', 'population', 'run_id']
     check_columns(expected_columns, data.columns)
     check_years(data, 'annual')
@@ -287,7 +287,7 @@ def _validate_structure(data, entity, location_id):
 
 
 def _validate_theoretical_minimum_risk_life_expectancy(data, entity, location_id):
-    del data, entity, location_id
+    del data, entity, location_id  # unused
     pass
 
 
--- a/src/vivarium_inputs/validation/raw.py
+++ b/src/vivarium_inputs/validation/raw.py
@@ -57,12 +57,12 @@ def validate_raw_data(data, entity, measure, location_id):
 def _check_sequela_metadata(entity, measure):
     if measure in ['incidence', 'prevalence']:
         if not entity[f'{measure}_exists']:
-            raise InvalidQueryError(f'{entity.name} does not have {measure} data')
+            raise InvalidQueryError(f'{entity.name} does not have {measure} data.')
         if not entity[f'{measure}_in_range']:
-            raise warnings.warn(f'{entity.name} has {measure} but its range is abnormal')
+            warnings.warn(f'{entity.name} has {measure} but its range is abnormal.')
     else:  # measure == 'disability_weight
-        if not entity.healthstate[f'{measure}_exist']:
-            raise InvalidQueryError(f'{entity.name} does not have {measure} data')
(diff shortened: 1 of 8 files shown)


src/vivarium_inputs/validation/raw.py
```diff
@@ -55,42 +56,80 @@ def validate_raw_data(data, entity, measure, location_id):
...
+    if measure == 'population_attributable_fraction':
+        _check_paf_types(entity)
+    else:
+        exists = entity[f'{mapping_names[measure]}_exists']
+        if exists is not None and not exists:
+            warnings.warn(f'{measure.capitalize()} data for risk factor {entity.name} may not exist for all locations.')
+
+        if measure == 'relative_risk' and exists and not entity.rr_in_range:
+            warnings.warn(f'{measure.capitalize()} data for risk factor {entity.name} may be outside '
+                          f'expected range >1.')
```
halfhorst: this is inclusive as I checked it

kfwilson: fixed!

src/vivarium_inputs/validation/raw.py
```diff
@@ -2,23 +2,29 @@
...
 import pandas as pd
+import numpy as np
+from typing import NamedTuple, Union
+
+from gbd_mapping import (ModelableEntity, Cause, Sequela, RiskFactor,
+                         Etiology, Covariate, CoverageGap)
 
 from vivarium_inputs.globals import (DRAW_COLUMNS, DEMOGRAPHIC_COLUMNS, METRICS,
                                      DataAbnormalError, InvalidQueryError, gbd)
+from vivarium_inputs.mapping_extension import HealthcareEntity, HealthTechnology
```
michaella21: Also AlternativeRiskFactor with others

src/vivarium_inputs/validation/raw.py
```diff
@@ -55,52 +61,85 @@ def validate_raw_data(data, entity, measure, location_id):
...
-        warnings.warn(f'{entity.name} has {measure} but its range is abnormal.')
+        if consistent and not entity[f"{measure}_aggregates"]:
+            warnings.warn(f"{children.capitalize()} {measure} data for cause {entity.name} may not correctly "
+                          f"aggregate up to the cause level in all locations. {children.capitalize()} models may not "
+                          f"be consistent with models for this cause.")
 
 
-def _check_covariate_metadata(entity, measure):
-    pass
+def check_risk_factor_metadata(entity: RiskFactor, measure: str):
```
michaella21: In the check_metadata, alternative risk factor also calls for the check_risk_factor_metadata. Technically it only has its matching risk_factor's metadata, but we still expect to pass it through check metadata

src/vivarium_inputs/validation/raw.py
kfwilson: done

src/vivarium_inputs/validation/raw.py
kfwilson: done

src/vivarium_inputs/validation/raw.py
```diff
@@ -2,23 +2,29 @@
...
+        'sequela': check_sequela_metadata,
+        'cause': check_cause_metadata,
+        'risk_factor': check_risk_factor_metadata,
+        'etiology': check_etiology_metadata,
+        'covariate': check_covariate_metadata,
+        'coverage_gap': check_coverage_gap_metadata,
+        'health_technology': check_health_technology_metadata,
+        'healthcare_entity': check_healthcare_entity_metadata,
+        'population': check_population_metadata,
+        'alternative_risk_factor': check_risk_factor_metadata,
```
kfwilson: @michaella21 mentioned that we are just copying over the mapping for the original risk factor to the alternative risk factor and changing only distribution and kind. It seems very weird to me to check the metadata for alternative risk factors if we are just copying the flags from the original risk factor. We may be giving warnings on data not existing/being out of range that are based on data from the original risk factor not the alternative risk factor. It seems like the best solution is to actually update all the flags for the alternative risk factor to reflect the data we actually have but if that's not workable, I'd vote for just passing on the risk factor metadata checks if the kind is alternative risk factor. Thoughts?

kfwilson: passing checks for alt risk factors

collijk: Make an explicit function and have it pass

src/vivarium_inputs/validation/raw.py
```diff
@@ -55,52 +61,88 @@ def validate_raw_data(data, entity, measure, location_id):
...
+
+        if measure == 'exposure' and entity.exposure_year_type in ('mix', 'incomplete'):
+            warnings.warn(f'{measure.capitalize()} data for risk factor {entity.name} may contain unexpected '
+                          f'or missing years.')
+
+    _warn_violated_restrictions(entity, measure)
+
+
+def check_etiology_metadata(entity: Etiology, measure: str):
+    del measure  # unused
```
collijk: Why delete it manually?

src/vivarium_inputs/validation/raw.py
```diff
@@ -55,52 +61,88 @@ def validate_raw_data(data, entity, measure, location_id):
...
+        if consistent and not entity[f"{measure}_aggregates"]:
+            warnings.warn(f"{children.capitalize()} {measure} data for cause {entity.name} may not correctly "
+                          f"aggregate up to the cause level in all locations. {children.capitalize()} models may not "
+                          f"be consistent with models for this cause.")
 
 
-def _check_covariate_metadata(entity, measure):
-    pass
+def check_risk_factor_metadata(entity: Union[AlternativeRiskFactor, RiskFactor], measure: str):
+    if entity.kind == 'alternative_risk_factor':
```
collijk: Remove this branch and replace with function

src/vivarium_inputs/validation/raw.py
kfwilson: I was reading a style guide the other day and it recommended doing this so you don't get linter warnings about unused args. I can take it out though since we haven't really done it anywhere else

src/vivarium_inputs/validation/raw.py
kfwilson: done!

src/vivarium_inputs/validation/raw.py
kfwilson: done!

src/vivarium_inputs/validation/raw.py
collijk: The typical convention (and one your linter will recognize) is to use ``_`` for an unused arg

kfwilson: Yeah but I don't love that in general because it breaks callers that pass by name. But I'll stick to that since that's what we've used elsewhere.

PR #72, opened 2019-01-18
bugfix/risk factor full names

kfwilson:
Didn't realize the alt risk factors use the risk factor constructor and aren't auto built so the changes to the risk factors to use full names broke these. Here's a fix.


kfwilson:
Oh no it was my fault for not catching this! I knew about the alt risk factors and just didn't think carefully about how they interacted with risk factors when I made the changes to rfs

PR #73, opened 2019-01-19
Feature/gbd 2017 raw validation cause

kfwilson:
Raw validators for the cause like measures + a bunch of helper functions + tests for those helper functions


src/vivarium_inputs/validation/raw.py
```diff
@@ -318,3 +418,434 @@ def _check_paf_types(entity):
...
+                                        'values for both males and females.')
+        else:  # check combined sex id
+            if not check_data_exist(data[data.sex_id == gbd.COMBINED], zeros_missing=True,
+                                    value_columns=value_columns, error=False):
+                raise DataAbnormalError('Data has no sex restrictions, but does not contain non-zero '
+                                        'values for both males and females.')
+
+
+def check_measure_id(data: pd.DataFrame, allowable_measures: List[str]):
+    """Check that data contains only a single measure id and that it is one of
```
michaella21: I think we do filter down both measure and metric in extract before going through validator. Do we have a reason do pass  allowable_measures as a list while metric id to be a single value in 828?

src/vivarium_inputs/validation/raw.py
```diff
@@ -318,3 +418,434 @@ def _check_paf_types(entity):
...
+
+    all_ages.sort()
+    restriction_ages.sort()
+    data_ages.sort()
+
+    invalid_age_ids = set(data_ages).difference(all_ages)
+    if invalid_age_ids:
+        raise DataAbnormalError(f'Data contains invalid age group ids: {invalid_age_ids}')
+
+    if data_ages == restriction_ages or data_ages == all_ages:
```
halfhorst: what if we get data that is supposed to be restricted but is not? i.e. restriction_ages is not empty but the data matches all_ages

src/vivarium_inputs/validation/raw.py
```diff
@@ -318,3 +418,434 @@ def _check_paf_types(entity):
...
+        return
+
+    # make sure data_ages is contiguous block
+    if all_ages[all_ages.index(data_ages[0]):all_ages.index(data_ages[-1])+1] != data_ages:
+        raise DataAbnormalError(f'Data contains a non-contiguous age groups: {data_ages}.')
+
+    if set(data_ages).issubset(restriction_ages):
```
halfhorst: FYI issubset is "inclusive." As it is it's fine, but if you were to change something above or move ordering it could warn unnecessarily. May be safer to explicitly use '<'

kfwilson: so I check earlier the case that the two are equal and return without warning to avoid warning on the inclusive case but probably better to use <

src/vivarium_inputs/validation/raw.py
kfwilson: that was planning ahead for exposure which can have three possible measures (although should only ever have 1 at a time)

tests/validation/test_raw.py
```diff
@@ -0,0 +1,234 @@
...
+
+from unittest import mock
+
+from vivarium_inputs.validation import raw
+from vivarium_inputs.globals import DataAbnormalError, DataNotExistError
+
+
+@mock.patch('vivarium_inputs.validation.raw.gbd.get_estimation_years')
+def test_check_years(mock_get_estimation_years):
+    mock_get_estimation_years.return_value = list(range(1990, 2015, 5)) + [2017]
```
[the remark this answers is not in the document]
halfhorst: one thing you could do is split the valid checks and the invalid checks and parametrize. When you parametrize you can add a little label, so when you run you get a nice description of what the test case is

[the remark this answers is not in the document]
collijk: Yeah, that's a good way to go.  You end up also having to pass in expected returns/error messages however.  That can get very verbose very quickly.  That's not necessarily bad, but it can obscure what's being tested.

src/vivarium_inputs/validation/raw.py
```diff
@@ -1,18 +1,26 @@
 from typing import List
 import warnings
-
+import operator
 import pandas as pd
 import numpy as np
-from typing import NamedTuple, Union
+from typing import NamedTuple, Union, List
```
halfhorst: this import is duplicated from line 1

michaella21:
Looks good! lots of useful helper functions and tests look good also! 👍

src/vivarium_inputs/validation/raw.py
```diff
@@ -634,11 +645,11 @@ def check_sex_ids(data: pd.DataFrame, male_expected: bool = True, female_expecte
         If data contains any sex ids that aren't valid GBD sex ids.
 
     """
-    valid_sex_ids = [gbd.MALE, gbd.FEMALE, gbd.COMBINED]
-    gbd_sex_ids = np.array(valid_sex_ids)[[male_expected, female_expected, combined_expected]]
+    valid_sex_ids = gbd.MALE + gbd.FEMALE + gbd.COMBINED  # these are single-item lists
```
collijk: the fact that they're single item lists is pretty dumb.  I wrote them that way when their only use was inside gbd.py

PR #76, opened 2019-01-23
extract risk-like/health system measures

michaella21:
extract data for risk-like measures and health systems

### DIFF
--- a/src/vivarium_inputs/extract.py
+++ b/src/vivarium_inputs/extract.py
@@ -79,7 +79,7 @@ def extract_exposure(entity, location_id: int) -> pd.DataFrame:
         proper_measure_id = set(data.measure_id).intersection(allowable_measures)
         if len(proper_measure_id) != 1:
             raise DataAbnormalError(f'Exposure data have {len(proper_measure_id)} measure id(s). Data should have'
-                                    f'exact one out of {allowable_measures} but came back with {proper_measure_id}')
+                                    f'exactly one id out of {allowable_measures} but came back with {proper_measure_id}.')
         else:
             data = data[data.measure_id == proper_measure_id.pop()]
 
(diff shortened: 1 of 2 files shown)


src/vivarium_inputs/extract.py
```diff
@@ -73,29 +73,26 @@ def extract_deaths(entity, location_id: int) -> pd.DataFrame:
 
 
 def extract_exposure(entity, location_id: int) -> pd.DataFrame:
-    if entity.kind == 'risk_factor' and entity.distribution == 'dichotomous':
+    if entity.kind == 'risk_factor':
         data = gbd.get_exposure(entity.gbd_id, location_id)
         if MEASURES['Proportion'] in data.measure_id.unique():
```
collijk: I think we should be slightly more careful about the measures check here. We should check that exactly one of the exposure measures are present, rather than the current strategy of picking the first one we like.

src/vivarium_inputs/extract.py
```diff
@@ -73,29 +73,26 @@ def extract_deaths(entity, location_id: int) -> pd.DataFrame:
...
         data = gbd.get_exposure(entity.gbd_id, location_id)
-        if MEASURES['Proportion'] in data.measure_id.unique():
-            data = data[data.measure_id == MEASURES['Proportion']]
-        elif MEASURES['Prevalence'] in data.measure_id.unique():  # Exposure comes from a cause model
-            data = data[data.measure_id == MEASURES['Prevalence']]
+        allowable_measures = [MEASURES['Proportion'], MEASURES['Continuous'], MEASURES['Prevalence']]
+        proper_measure_id = set(data.measure_id).intersection(allowable_measures)
+        if len(proper_measure_id) != 1:
+            raise DataAbnormalError(f'Exposure data have {len(proper_measure_id)} measure id(s). Data should have'
+                                    f'exact one out of {allowable_measures} but came back with {proper_measure_id}')
```
kfwilson: period please :)

michaella21: yep, easy to miss when it ends with the right bracket :( I added it!

PR #77, opened 2019-01-24
cause disability weight is zero if no sequela

halfhorst:
if a cause has no sequelae, return 0.0 as its disability weight

### DIFF
--- a/src/vivarium_inputs/core.py
+++ b/src/vivarium_inputs/core.py
@@ -1,12 +1,14 @@
 from collections import namedtuple
 from typing import Union
+import itertools
 
 from gbd_mapping import Cause, RiskFactor, Sequela, Covariate
 import pandas as pd
 import numpy as np
 
+from vivarium_gbd_access import gbd
 from vivarium_inputs import utilities, extract
-from .globals import InvalidQueryError, DEMOGRAPHIC_COLUMNS, DRAW_COLUMNS
+from vivarium_inputs.globals import InvalidQueryError, DEMOGRAPHIC_COLUMNS, DRAW_COLUMNS
 
 
 def get_data(entity, measure: str, location: str):
@@ -79,14 +81,24 @@ def get_birth_prevalence(entity: Union[Cause, Sequela], location_id: int) -> pd.
 
 def get_disability_weight(entity: Union[Cause, Sequela], location_id: int) -> pd.DataFrame:
     if entity.kind == 'cause':
-        partial_weights = []
-        for sequela in entity.sequelae:
-            prevalence = get_prevalence(sequela, location_id).set_index(list(DEMOGRAPHIC_COLUMNS) + ['draw'])
-            disability = get_disability_weight(sequela, location_id)
-            disability['location_id'] = location_id
-            disability = disability.set_index(list(DEMOGRAPHIC_COLUMNS) + ['draw'])
-            partial_weights.append(prevalence*disability)
-        data = sum(partial_weights).reset_index()
+        if entity.sequelae:
+            partial_weights = []
+            for sequela in entity.sequelae:
+                prevalence = get_prevalence(sequela, location_id).set_index(list(DEMOGRAPHIC_COLUMNS) + ['draw'])
+                disability = get_disability_weight(sequela, location_id)
+                disability['location_id'] = location_id
+                disability = disability.set_index(list(DEMOGRAPHIC_COLUMNS) + ['draw'])
+                partial_weights.append(prevalence*disability)
+            data = sum(partial_weights).reset_index()
+        else:  # assume disability weight is zero if no sequela
+            sex_id = [1, 2]
+            age_group_id = gbd.get_age_group_id()
+            location_id = [location_id]
+            year_id = range(1990, 2018)
+            draw = range(1000)
+            value = [0.0]
+            data = pd.DataFrame(data=list(itertools.product(sex_id, age_group_id, location_id, year_id, draw, value)),
+                                columns=["sex_id", "age_group_id", "location_id", "year_id", "draw", "value"])
     else:  # entity.kind == 'sequela'
         data = extract.extract_data(entity, 'disability_weight', location_id)
         data = utilities.normalize(data)
--- a/src/vivarium_inputs/core.py
+++ b/src/vivarium_inputs/core.py
@@ -1,12 +1,10 @@
 from collections import namedtuple
 from typing import Union
-import itertools
 
-from gbd_mapping import Cause, RiskFactor, Sequela, Covariate
+from gbd_mapping import Cause, Sequela
 import pandas as pd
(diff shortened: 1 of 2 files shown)


src/vivarium_inputs/core.py
```diff
@@ -79,14 +81,24 @@ def get_birth_prevalence(entity: Union[Cause, Sequela], location_id: int) -> pd.
...
+            partial_weights = []
+            for sequela in entity.sequelae:
+                prevalence = get_prevalence(sequela, location_id).set_index(list(DEMOGRAPHIC_COLUMNS) + ['draw'])
+                disability = get_disability_weight(sequela, location_id)
+                disability['location_id'] = location_id
+                disability = disability.set_index(list(DEMOGRAPHIC_COLUMNS) + ['draw'])
+                partial_weights.append(prevalence*disability)
+            data = sum(partial_weights).reset_index()
+        else:  # assume disability weight is zero if no sequela
+            sex_id = [1, 2]
```
kfwilson: maybe use the constants from vivarium_gbd_access.gbd

halfhorst: is there one from year, too?

kfwilson: I don't think so. there's get_gbd_estimation_years but that gives you back the binned years and this needs annual

src/vivarium_inputs/core.py
```diff
@@ -79,14 +81,24 @@ def get_birth_prevalence(entity: Union[Cause, Sequela], location_id: int) -> pd.
...
+            partial_weights = []
+            for sequela in entity.sequelae:
+                prevalence = get_prevalence(sequela, location_id).set_index(list(DEMOGRAPHIC_COLUMNS) + ['draw'])
+                disability = get_disability_weight(sequela, location_id)
+                disability['location_id'] = location_id
+                disability = disability.set_index(list(DEMOGRAPHIC_COLUMNS) + ['draw'])
+                partial_weights.append(prevalence*disability)
+            data = sum(partial_weights).reset_index()
+        else:  # assume disability weight is zero if no sequela
+            sex_id = gbd.MALE + gbd.FEMALE
```
collijk: Add a default `False` draw argument to `utilities.get_demographic_dimensions` and use that to construct your key column block.

src/vivarium_inputs/core.py
```diff
@@ -79,14 +81,24 @@ def get_birth_prevalence(entity: Union[Cause, Sequela], location_id: int) -> pd.
...
     if entity.kind == 'cause':
-        partial_weights = []
-        for sequela in entity.sequelae:
-            prevalence = get_prevalence(sequela, location_id).set_index(list(DEMOGRAPHIC_COLUMNS) + ['draw'])
-            disability = get_disability_weight(sequela, location_id)
-            disability['location_id'] = location_id
-            disability = disability.set_index(list(DEMOGRAPHIC_COLUMNS) + ['draw'])
-            partial_weights.append(prevalence*disability)
-        data = sum(partial_weights).reset_index()
+        if entity.sequelae:
```
collijk: You can remove the whole branch here if you initialize your data with all zeros, then collect the cumulative sum in the loop.

src/vivarium_inputs/core.py
```diff
@@ -79,14 +81,24 @@ def get_birth_prevalence(entity: Union[Cause, Sequela], location_id: int) -> pd.
...
+            partial_weights = []
+            for sequela in entity.sequelae:
+                prevalence = get_prevalence(sequela, location_id).set_index(list(DEMOGRAPHIC_COLUMNS) + ['draw'])
+                disability = get_disability_weight(sequela, location_id)
+                disability['location_id'] = location_id
```
collijk: The sequela branch of this function should add the location id to the disability weight dataframe (if it's not already).  The return from this function should always have the same shape.

src/vivarium_inputs/core.py
```diff
@@ -79,14 +81,24 @@ def get_birth_prevalence(entity: Union[Cause, Sequela], location_id: int) -> pd.
...
+            age_group_id = gbd.get_age_group_id()
+            location_id = [location_id]
+            year_id = range(1990, 2018)
+            draw = range(1000)
+            value = [0.0]
+            data = pd.DataFrame(data=list(itertools.product(sex_id, age_group_id, location_id, year_id, draw, value)),
+                                columns=["sex_id", "age_group_id", "location_id", "year_id", "draw", "value"])
     else:  # entity.kind == 'sequela'
         data = extract.extract_data(entity, 'disability_weight', location_id)
         data = utilities.normalize(data)
```
collijk: We should probably make normalize and reshape idempotent and pull them out of the branch statement.  Not required for this PR.  Just a note.

src/vivarium_inputs/core.py
collijk: Add the constants to that function for sexes as well.

src/vivarium_inputs/core.py
halfhorst: yup, this is just for the merge

src/vivarium_inputs/core.py
halfhorst: normalize is idempotent. I made reshape idempotent by checking for the column that results from the reshape and not reshaping if it is present

src/vivarium_inputs/core.py
collijk: You now no longer need the `if entity.sequelae` check.  `entity.sequela` definitely exists since it's a cause, and it's an empty iterable if no sequelae are present.

src/vivarium_inputs/utilities.py
```diff
@@ -178,8 +185,9 @@ def normalize_age(data: pd.DataFrame, fill_value) -> pd.DataFrame:
 
 
 def reshape(data: pd.DataFrame, to_keep=DEMOGRAPHIC_COLUMNS) -> pd.DataFrame:
-    data = pd.melt(data, id_vars=to_keep, value_vars=DRAW_COLUMNS, var_name='draw')
-    data["draw"] = data.draw.str.partition("_")[2].astype(int)
+    if 'draw' not in data.columns:
```
collijk: Revert this for now.  It's not nearly careful enough.  If we're going to do this, we'd structure it to be able to move normalize and reshape up into `core.get_data` instead of having them called in each function.

src/vivarium_inputs/core.py
collijk: :+1:

src/vivarium_inputs/core.py
collijk: I am wrong.  It's none.  It should be an empty iterable though, probably.

src/vivarium_inputs/utilities.py
halfhorst: :+1:

PR #78, opened 2019-01-24
Feature/gbd 2017 raw validation risk exposure

kfwilson:


src/vivarium_inputs/validation/raw.py
```diff
@@ -318,16 +321,55 @@ def _validate_deaths(data: pd.DataFrame, entity: Cause, location_id: int):
...
+
+    cats.apply(check_age_group_ids, age_start, age_end)
+    cats.apply(check_sex_ids, male_expected, female_expected)
+
+    if not check_years(data, 'annual', error=False) and not check_years(data, 'binned', error=False):
+        raise DataAbnormalError(f'Exposure data for {entity.kind} {entity.name} contains a year range '
+                                f'that is neither annual nor binned.')
     check_location(data, location_id)
 
+    if entity.kind == 'risk_factor':
```
collijk: You have two if blocks for risk factor?

src/vivarium_inputs/validation/raw.py
```diff
@@ -318,16 +321,55 @@ def _validate_deaths(data: pd.DataFrame, entity: Cause, location_id: int):
...
+        cats.apply(check_sex_restrictions, entity.restrictions.male_only, entity.restrictions.female_only)
+
+    if entity.distribution in ('ensemble', 'lognormal', 'normal'):  # continuous
+        if entity.tmred.inverted:
+            check_value_columns_boundary(data, entity.tmred.max, 'upper',
+                                         value_columns=DRAW_COLUMNS, inclusive=True, error=False)
+        else:
+            check_value_columns_boundary(data, entity.tmred.min, 'lower',
+                                         value_columns=DRAW_COLUMNS, inclusive=True, error=False)
+    else:  # categorical
```
michaella21: exposure can be risk factor/coverage gap/ alternative risk factor(this is currently all continuous)

src/vivarium_inputs/validation/raw.py
kfwilson: Yes because I was trying to keep the same order of checks in every function and the parameters for checking age, sex ids are different for risk factors because they have the same restrictions (first block) but then risk factor, coverage gap, alt risk factor should all have years, loc checked and then only risk factors should check age/sex restrictions in draw columns (second block).

src/vivarium_inputs/validation/raw.py
kfwilson: hmm so we won't have tmred to use as the boundary to check for continuous coverage gaps and alt risk factors. Any suggestions on what boundary values to use for those entities @collijk?

michaella21: yeah, that seems problematic here. I understand why you split in this way but there are very limited things that we can check for the alt risk factors and coverage gaps. (well we can `optimistically` think that we won't need to validate coverage gaps once we can check all of these during the aux validation, some day... :( )

src/vivarium_inputs/validation/raw.py
kfwilson: I reordered all the checks so that I could have single risk factor block that also now includes the boundary checks for continuous distributions since @michaella21 pointed out that we only have the tmred info for risk factors

src/vivarium_inputs/validation/raw.py
kfwilson: ok I moved that check for continuous dist into the risk factor only block. that means now that we have no boundary checks on continuous dists from coverage gaps and alt risk factors which I don't like, but I don't know what reasonable boundaries are for those

PR #79, opened 2019-01-25
Feature/gbd 2017 core risk like

michaella21:
risk-like measures and cost/utilization core parts.


src/vivarium_inputs/core.py
```diff
@@ -128,39 +128,30 @@ def _get_deaths(entity: Cause, location_id: int) -> pd.DataFrame:
...
-    elif entity.kind == 'alternative_risk_factor':
-        data = extract.extract_data(entity, 'exposure', location_id)
-        # FIXME: me_id is usually nan which breaks year interpolation. This is a stupid hack.
-        data = data.drop('modelable_entity_id', 'columns')
-        data = utilities.normalize(data, fill_value=0)
-        data = utilities.reshape(data)
-    else:
-        raise NotImplementedError()
+    data = extract.extract_data(entity, 'exposure', location_id)
+    categorical = True if entity.distribution in ['dichotomous', 'polytomous', 'ordered_polytomous'] else False
```
kfwilson: I think this should be ['dichotomous', 'unordered_polytomous', 'ordered_polytomous']

src/vivarium_inputs/core.py
```diff
@@ -128,39 +128,30 @@ def _get_deaths(entity: Cause, location_id: int) -> pd.DataFrame:
...
+    if set(data.measure_id) == {MEASURES['Prevalence']}:
+        data = _convert_prevalence_to_proportion(data)
+    data = utilities.normalize(data, fill_value=0, categorical=categorical)
+    data = utilities.reshape(data, to_keep=list(DEMOGRAPHIC_COLUMNS) + ['parameter'])
```
kfwilson: you shouldn't have to do list(DEMOGRAPHIC_COLUMNS) - I changed DEMOGRAPHIC_COLS to a list in my cause validators PR

src/vivarium_inputs/core.py
```diff
@@ -169,64 +160,45 @@ def get_exposure_standard_deviation(entity, location_id):
...
-    else:
-        raise NotImplementedError()
+    categorical = True if entity.distribution in ['dichotomous', 'polytomous', 'ordered_polytomous'] else False
+    result = []
+    for affected_entity in data.affected_entity.unique():
+        df = data[data.affected_entity == affected_entity]
+        df = utilities.normalize(df, fill_value=1, categorical=categorical)
+        result.append(df)
+    data = pd.concat(result)
+    data = utilities.reshape(data, to_keep=list(DEMOGRAPHIC_COLUMNS)
```
kfwilson: shouldn't need to coerce DEMOGRAPHIC_COLS to list here too

src/vivarium_inputs/core.py
```diff
@@ -169,64 +160,45 @@ def get_exposure_standard_deviation(entity, location_id):
...
     if entity.kind == 'risk_factor':
-        data = extract.extract_data(entity, 'relative_risk', location_id)
         data = utilities.convert_affected_entity(data, 'cause_id')
         data['affected_measure'] = 'incidence_rate'
-        data = utilities.normalize(data, fill_value=1)
-        data = utilities.reshape(data, to_keep=list(DEMOGRAPHIC_COLUMNS)
-                                               + ['affected_entity', 'affected_measure', 'parameter'])
-    elif entity.kind == 'coverage_gap':
-        data = extract.extract_data(entity, 'relative_risk', location_id)
+    else:
```
kfwilson: so alternative risk factors also always target exposure params?

src/vivarium_inputs/core.py
michaella21: oops :( I thought that that change was a part of risk validator PR

src/vivarium_inputs/core.py
```diff
@@ -169,64 +160,45 @@ def get_exposure_standard_deviation(entity, location_id):
...
-            pafs.append(paf)
-        data = pd.concat(pafs)
-        data['location_id'] = location_id
-    else:
-        raise NotImplementedError()
+    data = extract.extract_data(entity, 'population_attributable_fraction', location_id)
+    data = utilities.convert_affected_entity(data, 'cause_id')
+    data['affected_measure'] = 'incidence_rate'
+    data = utilities.normalize(data, fill_value=0)
+    data = utilities.reshape(data, to_keep=list(DEMOGRAPHIC_COLUMNS) + ['affected_entity', 'affected_measure'])
```
kfwilson: same comment re list(DEMOGRAPHIC_COLS)

src/vivarium_inputs/core.py
```diff
@@ -245,15 +217,17 @@ def get_estimate(entity, location_id):
 
 
 def get_cost(entity, location_id):
-    if entity.kind == 'healthcare_entity':
-        raise NotImplementedError()
-    elif entity.kind == 'health_technology':
-        raise NotImplementedError()
+    data = extract.extract_data(entity, 'cost', location_id)
+    data = utilities.normalize(data, fill_value=0)
+    data = utilities.reshape(data, to_keep=list(DEMOGRAPHIC_COLUMNS) + [entity.kind])
```
kfwilson: no need for list

src/vivarium_inputs/core.py
michaella21: no, core only get risk factor and coverage gap for RR, at least that's how it is now.

src/vivarium_inputs/utilities.py
```diff
@@ -86,7 +86,21 @@ def scrub_affected_entity(data):
 ###############################################################
 
 
-def normalize(data: pd.DataFrame, fill_value=None) -> pd.DataFrame:
+def normalize(data: pd.DataFrame, fill_value=None, categorical=False):
+    if categorical:
+        cats = data.parameter.unique()
+        df = []
+        for cat in cats:
```
kfwilson: I think you could do this with a groupby + apply rather than looping over each cat?

src/vivarium_inputs/core.py
kfwilson: oh right! Can we add a comment by the else block then just to remind us that we're always saying coverage gaps target exposure params?

src/vivarium_inputs/utilities.py
```diff
@@ -121,6 +135,9 @@ def normalize_sex(data: pd.DataFrame, fill_value) -> pd.DataFrame:
 
 def normalize_year(data: pd.DataFrame) -> pd.DataFrame:
     years = {'annual': list(range(1990, 2018)), 'binned': gbd.get_estimation_years()}
+    if set(data.year_id)=={1}:
```
kfwilson: this is from aux data right? I don't love having this random 1 here. Maybe a comment at least explaining why we are checking for 1?

src/vivarium_inputs/utilities.py
```diff
@@ -142,8 +159,11 @@ def normalize_year(data: pd.DataFrame) -> pd.DataFrame:
 def interpolate_year(data):
     # Hide the central comp dependency unless required.
     from core_maths.interpolate import pchip_interpolate
-    id_cols = data.columns.difference(DRAW_COLUMNS)
-    fillin_data = pchip_interpolate(data, id_cols, DRAW_COLUMNS)
+    nan_cols = data.columns[data.isna().any()].tolist()
```
kfwilson: probably won't make a difference but you don't technically need to check that there are no NaNs in DRAW_COLUMNS - I'm breaking in raw validation if there's any NaNs in any draw col

src/vivarium_inputs/utilities.py
```diff
@@ -142,8 +159,11 @@ def normalize_year(data: pd.DataFrame) -> pd.DataFrame:
...
+    non_nan_cols = set(data.columns).difference(nan_cols)
+    id_cols =non_nan_cols.difference(DRAW_COLUMNS)
+    fillin_data = pchip_interpolate(data, list(id_cols), DRAW_COLUMNS)
+    fillin_data = pd.concat([fillin_data, pd.DataFrame(columns=nan_cols)])
```
kfwilson: maybe I'm missing something but aren't we possibly losing data here? You're pulling out any column that has even a single NaN in it and then putting back those columns here but they're all empty so you've potentially lost the 99.99% of the column that wasn't NaN - do we not care about that?

src/vivarium_inputs/core.py
```diff
@@ -169,64 +157,44 @@ def get_exposure_standard_deviation(entity, location_id):
...
-        affected_entities = rrs.affected_entity.unique()
-        pafs = []
-        for affected_entity in affected_entities:
-            paf = utilities.compute_categorical_paf(rrs, e, affected_entity)
-            pafs.append(paf)
-        data = pd.concat(pafs)
-        data['location_id'] = location_id
-    else:
-        raise NotImplementedError()
+    data = extract.extract_data(entity, 'population_attributable_fraction', location_id)
```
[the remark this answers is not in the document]
michaella21: we already drop measure id/metric id to be YLDs/percent in the extract so core does not filter or drop anything. If we need a measure id YLL, we need to change the extract part first and also pass the measure id checks which wants data to have a single measure id

src/vivarium_inputs/core.py
```diff
@@ -129,37 +129,18 @@ def _get_deaths(entity: Cause, location_id: int) -> pd.DataFrame:
...
-        data = data.drop('modelable_entity_id', 'columns')
-        data = utilities.normalize(data, fill_value=0)
-        data = utilities.reshape(data)
+    data = extract.extract_data(entity, 'exposure', location_id)
+    data = data.drop('modelable_entity_id', 'columns')
+    if entity.distribution in ['dichotomous', 'ordered_polytomous', 'unordered_polytomous']:
+        data.groupby('parameter').apply(lambda df: utilities.normalize(df, fill_value=0))
     else:
-        raise NotImplementedError()
+        data = utilities.normalize(data, fill_value=0)
```
kfwilson: don't the continuous risks also have parameter in the data? i.e., couldn't you always do the groupby call because in the categorical case, you'll end up with different groups and in the continuous case, you'll end up with the original dataframe because there's only one parameter?

src/vivarium_inputs/core.py
```diff
@@ -169,64 +150,47 @@ def get_exposure_standard_deviation(entity, location_id):
...
-        data = utilities.reshape(data, to_keep=list(DEMOGRAPHIC_COLUMNS)
-                                               + ['affected_entity', 'affected_measure', 'parameter'])
-
-    else:
-        raise NotImplementedError()
+    result = []
+    for affected_entity in data.affected_entity.unique():
+        df = data[data.affected_entity == affected_entity]
+        if entity.distribution in ['dichotomous', 'ordered_polytomous', 'unordered_polytomous']:
+            df.groupby('parameter').apply(lambda d: utilities.normalize(d, fill_value=1))
```
kfwilson: same comment as above re always being able to do the groupby call regardless of distribution type

src/vivarium_inputs/core.py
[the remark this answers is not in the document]
kfwilson: I can easily change the measure id validator to allow multiple measure ids - I'm just not sure which measures we would allow to have multiple ids?

src/vivarium_inputs/utilities.py
```diff
@@ -142,8 +143,8 @@ def normalize_year(data: pd.DataFrame) -> pd.DataFrame:
 def interpolate_year(data):
     # Hide the central comp dependency unless required.
     from core_maths.interpolate import pchip_interpolate
-    id_cols = data.columns.difference(DRAW_COLUMNS)
-    fillin_data = pchip_interpolate(data, id_cols, DRAW_COLUMNS)
+    id_cols = list(set(data.columns).difference(DRAW_COLUMNS))
+    fillin_data = pchip_interpolate(data, list(id_cols), DRAW_COLUMNS)
```
kfwilson: you're coercing id_cols to a list twice

src/vivarium_inputs/core.py
michaella21: yep we can do that and it won't change anything on continuous case

src/vivarium_inputs/core.py
```diff
@@ -173,64 +152,43 @@ def get_exposure_standard_deviation(entity, location_id):
...
+    data = pd.melt(data, id_vars=id_cols, value_vars=distribution_cols, var_name='parameter')
     return data
 
 
 def get_relative_risk(entity, location_id):
+    data = extract.extract_data(entity, 'relative_risk', location_id)
     if entity.kind == 'risk_factor':
-        data = extract.extract_data(entity, 'relative_risk', location_id)
         data = utilities.convert_affected_entity(data, 'cause_id')
         data['affected_measure'] = 'incidence_rate'
```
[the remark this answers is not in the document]
michaella21: yeah, is https://github.com/ihmeuw/vivarium_inputs/blob/develop/src/vivarium_inputs/core.py#L267 checked at the raw validator or should we raise in core?

src/vivarium_inputs/core.py
```diff
@@ -133,37 +134,15 @@ def _get_deaths(entity: Cause, location_id: int) -> pd.DataFrame:
...
-        data = extract.extract_data(entity, 'exposure', location_id)
-        # FIXME: me_id is usually nan which breaks year interpolation. This is a stupid hack.
-        data = data.drop('modelable_entity_id', 'columns')
-        data = utilities.normalize(data, fill_value=0)
-        data = utilities.reshape(data)
-    else:
-        raise NotImplementedError()
+    data = extract.extract_data(entity, 'exposure', location_id)
+    data = data.drop('modelable_entity_id', 'columns')
+    data.groupby('parameter').apply(lambda df: utilities.normalize(df, fill_value=0))
```
kfwilson: does this happen in place? I would think you would want 
`data = data.groupby('parameter')....`

src/vivarium_inputs/core.py
```diff
@@ -133,37 +134,15 @@ def _get_deaths(entity: Cause, location_id: int) -> pd.DataFrame:
 
 
 def get_exposure(entity, location_id):
```
kfwilson: can we add type annotation for this + the methods below like we have above?

src/vivarium_inputs/core.py
```diff
@@ -173,64 +152,48 @@ def get_exposure_standard_deviation(entity, location_id):
...
-        data = utilities.normalize(data, fill_value=1)
-        data = utilities.reshape(data, to_keep=list(DEMOGRAPHIC_COLUMNS)
-                                               + ['affected_entity', 'affected_measure', 'parameter'])
-
-    else:
-        raise NotImplementedError()
+    result = []
+    for affected_entity in data.affected_entity.unique():
+        df = data[data.affected_entity == affected_entity]
+        df.groupby('parameter').apply(lambda d: utilities.normalize(d, fill_value=1))
```
kfwilson: I think you want `df = df.groupby....` here too?

src/vivarium_inputs/core.py
michaella21: of course it doesn't happen in place :(

PR #80, opened 2019-01-25
Feature/gbd 2017 raw validation risk exp sd weights

kfwilson:
validators for exposure standard deviation + exposure distribution weights

### DIFF
--- a/src/vivarium_inputs/validation/raw.py
+++ b/src/vivarium_inputs/validation/raw.py
@@ -372,10 +372,32 @@ def _validate_exposure(data: pd.DataFrame, entity: Union[RiskFactor, CoverageGap
 
 
 def _validate_exposure_standard_deviation(data, entity, location_id):
+    check_data_exist(data, zeros_missing=True)
 
-    expected_columns = ('rei_id', 'modelable_entity_id', 'measure_id', 'metric_id') + DEMOGRAPHIC_COLUMNS + DRAW_COLUMNS
+    expected_columns = ['rei_id', 'modelable_entity_id', 'measure_id',
+                        'metric_id'] + DEMOGRAPHIC_COLUMNS + DRAW_COLUMNS
     check_columns(expected_columns, data.columns)
+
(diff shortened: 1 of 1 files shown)


src/vivarium_inputs/validation/raw.py
```diff
@@ -365,24 +365,70 @@ def _validate_exposure(data: pd.DataFrame, entity: Union[RiskFactor, CoverageGap
...
     check_columns(key_cols + distribution_cols, data.columns)
 
+    if set(data.measure) != {'ensemble_distribution_weight'}:
+        raise DataAbnormalError(f'Exposure distribution weight data for {entity.kind} {entity.name} '
+                                f'contains abnormal measure values.')
+
+    check_location(data, location_id)
+
+    all_ages_age_group_id = 22
+    if data.age_group_id.unique() != all_ages_age_group_id:
```
halfhorst: do you need to make sure you will only have one age_group_id in the data or are you sure

kfwilson: hmm good point. there should only be one age group id but I should error if not. Fixed now

src/vivarium_inputs/validation/raw.py
```diff
@@ -365,24 +365,70 @@ def _validate_exposure(data: pd.DataFrame, entity: Union[RiskFactor, CoverageGap
         check_value_columns_boundary(data, 1, 'upper', value_columns=DRAW_COLUMNS, inclusive=True, error=True)
 
         g = data.groupby(DEMOGRAPHIC_COLUMNS)[DRAW_COLUMNS].sum()
-        if not np.all(g[DRAW_COLUMNS] == 1):
+        if not np.all(np.isclose(g[DRAW_COLUMNS], 1.0)):
```
michaella21: can't we do just np.allclose(g, 1.0) since you only have draw_columns in g?

kfwilson: yep. good call!

PR #81, opened 2019-01-25
Feature/gbd 2017 catch error extract

michaella21:
catch the expected exceptions that we found from the survey

### DIFF
--- a/src/vivarium_inputs/extract.py
+++ b/src/vivarium_inputs/extract.py
@@ -1,6 +1,9 @@
 import pandas as pd
 
-from .globals import gbd, METRICS, MEASURES, DataAbnormalError
+from get_draws.api import EmptyDataFrameException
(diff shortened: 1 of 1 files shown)


src/vivarium_inputs/extract.py
```diff
@@ -31,7 +34,17 @@ def extract_data(entity, measure: str, location_id: int) -> pd.DataFrame:
     }
 
     validation.check_metadata(entity, measure)
-    data = extractors[measure](entity, location_id)
+
+    try:
+        data = extractors[measure](entity, location_id)
+    except (ValueError, AssertionError, EmptyDataFrameException, NoBestVersionError) as e:
```
kfwilson: I think you also want InputsException - see https://stash.ihme.washington.edu/projects/CSTE/repos/gbd_survey/browse/gbd_survey/utilities.py#110 
which comes from pulling birth prev when it doesn't exist

PR #82, opened 2019-01-25
Feature/gbd 2017 raw validation rel risk

kfwilson:
validator for relative risk + small addition to exp b/c I left out the restrictions check when I re-ordered it before

Also fixes an issue where some risk factors have yld_age or yll_age but not both


src/vivarium_inputs/validation/raw.py
```diff
@@ -431,12 +434,52 @@ def _validate_exposure_distribution_weights(data, entity, location_id):
...
+    max_val = MAX_CATEG_REL_RISK if entity.distribution in ('ensemble', 'lognormal', 'normal') else MAX_CONT_REL_RISK
+    check_value_columns_boundary(data, max_val, 'upper', value_columns=DRAW_COLUMNS, inclusive=True, error=False)
+
+    valid_morb_mort_values = {0, 1}
+    for m in ['morbidity', 'mortality']:
+        if not set(data[m]).issubset(valid_morb_mort_values):
+            raise DataAbnormalError(f'Relative risk data for {entity.kind} {entity.name} '
+                                    f'contains values for {m} outside the expected {valid_morb_mort_values}.')
+
+    if (((data.morbidity == 1) & (data.mortality == 0)).any()
```
kfwilson: @collijk do we still want to make this check? I can't remember why we put it in but this means that the 4 risk factors that are yld_only will always break here and I can't remember if that was what we wanted or not.

collijk: Implement check

PR #85, opened 2019-01-29
Feature/gbd 2017 raw validation cost

kfwilson:
metadata check + validator for cost 

(plus two parentheses in error messages in other validators that I realized were missing)

### DIFF
--- a/src/vivarium_inputs/validation/raw.py
+++ b/src/vivarium_inputs/validation/raw.py
@@ -151,8 +151,7 @@ def check_coverage_gap_metadata(entity: CoverageGap, measure: str):
 
 def check_health_technology_metadata(entity: HealthTechnology, measure: str):
     if measure == 'cost':
-        warnings.warn(f'Cost data for {entity.kind} {entity.name} is constantly extrapolated outside of '
-                      f'years [1990, 2017].')
+        warnings.warn(f'Cost data for {entity.kind} {entity.name} does not vary by year.')
 
 
 def check_healthcare_entity_metadata(entity: HealthcareEntity, measure: str):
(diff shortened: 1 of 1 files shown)


src/vivarium_inputs/validation/raw.py
```diff
@@ -462,8 +465,26 @@ def _validate_estimate(data, entity, location_id):
...
+    check_columns(expected_columns, data.columns)
+
+    if set(data.measure) != {'cost'}:
+        raise DataAbnormalError(f'Cost data for {entity.kind} {entity.name} contains '
+                                f'measures beyond the expected cost.')
+
+    check_years(data, 'annual')
+    check_location(data, location_id)
+
+    all_ages_age_group_id = 22
```
collijk: This gets used in enough places that we should pull it out as a global variable (probably do it as its own PR)

src/vivarium_inputs/validation/raw.py
```diff
@@ -150,11 +150,14 @@ def check_coverage_gap_metadata(entity: CoverageGap, measure: str):
...
 def check_health_technology_metadata(entity: HealthTechnology, measure: str):
-    raise NotImplementedError()
+    if measure == 'cost':
+        warnings.warn(f'Cost data for {entity.kind} {entity.name} does not vary by year.')
 
 
 def check_healthcare_entity_metadata(entity: HealthcareEntity, measure: str):
-    raise NotImplementedError()
+    if measure == 'cost':
+        warnings.warn(f'Cost data for {entity.kind} {entity.name} is constantly extrapolated outside of '
```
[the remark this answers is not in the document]
kfwilson: fixed

src/vivarium_inputs/validation/raw.py
kfwilson: ok. I think its use is scattered across a couple open PRs but once those are all merged, I'll make a single PR that pulls this out in a global variable.

src/vivarium_inputs/validation/raw.py
```diff
@@ -150,11 +150,14 @@ def check_coverage_gap_metadata(entity: CoverageGap, measure: str):
...
-    raise NotImplementedError()
+    if measure == 'cost':
+        warnings.warn(f'Cost data for {entity.kind} {entity.name} does not vary by year.')
 
 
 def check_healthcare_entity_metadata(entity: HealthcareEntity, measure: str):
-    raise NotImplementedError()
+    if measure == 'cost':
+        warnings.warn(f'2017 cost data for {entity.kind} {entity.name} is duplicated from 2016 data, and all data '
+                      f'before 1990 is backfilled from 1990 data.')
```
michaella21: maybe before 1995?

kfwilson: good catch. fixed now!

PR #87, opened 2019-01-29
validator for pop structure

kfwilson:


src/vivarium_inputs/validation/raw.py
```diff
@@ -470,12 +470,20 @@ def _validate_utilization(data, entity, location_id):
     raise NotImplementedError()
 
 
-def _validate_structure(data, entity, location_id):
+def _validate_structure(data: pd.DataFrame, entity: NamedTuple, location_id: int):
+    check_data_exist(data, zeros_missing=True, value_columns=['population'])
+
     expected_columns = ['age_group_id', 'location_id', 'year_id', 'sex_id', 'population', 'run_id']
```
collijk: Do we know what run id is?  I didn't investigate it when I was trying to make models work (or in past iterations).

src/vivarium_inputs/validation/raw.py
```diff
@@ -470,12 +470,20 @@ def _validate_utilization(data, entity, location_id):
...
     check_columns(expected_columns, data.columns)
+
     check_years(data, 'annual')
     check_location(data, location_id)
 
+    check_age_group_ids(data, None, None)
+    check_sex_ids(data, male_expected=True, female_expected=True, combined_expected=True)
+
+    check_value_columns_boundary(data, 0, 'lower', value_columns=['population'], inclusive=True, error=True)
```
collijk: Let's make a dumb maximum here.  Population of china and india are each around 1.5 billion.  we have 20-ish age groups and 2 sexes.  Let's say not population value should be bigger than 50 million.

collijk: Ah, 100 million.  Since we have a combined category for sexes

src/vivarium_inputs/validation/raw.py
kfwilson: So there's nothing in the db_queries get_pop documentation about run_id but I pulled a bunch of different locations at various levels in the loc hierarchy and it's always the same (117). Should I investigate further?

src/vivarium_inputs/validation/raw.py
kfwilson: done!

PR #89, opened 2019-01-29
Feature/sim validation checks

halfhorst:
this encompasses age and sex restriction checks and tests for the sim column/restriction checks.

Once this is merged I will open a PR with the cause-like validators so that the diffs look nice.


src/vivarium_inputs/validation/sim.py
```diff
@@ -142,77 +146,106 @@ def _validate_demographic_dimensions(data, entity, location):
...
     _validate_year_columns(data)
-    _validate_value_column(data)
 
 
-def _validate_draw_column(data):
+def _validate_draw_column(data: pd.DataFrame):
     if 'draw' not in data.columns:
         raise DataFormattingError('Draw column name improperly specified.')
 
     if list(data['draw'].unique()) != list(range(1000)):
```
kfwilson: do we want to use sets here so it doesn't require that the draw column be sorted? Or is it guaranteed that it will be sorted?

src/vivarium_inputs/validation/sim.py
```diff
@@ -142,77 +146,106 @@ def _validate_demographic_dimensions(data, entity, location):
...
+
+def _translate_age_restrictions(ids: Sequence[int]) -> (float, float):
+    age_bins = utilities.get_age_bins()
+    minimum = age_bins.loc[age_bins.age_group_id.isin(ids), 'age_group_start'].min()
+    maximum = age_bins.loc[age_bins.age_group_id.isin(ids), 'age_group_end'].max()
+
+    return minimum, maximum
+
+
+def _check_age_restrictions(data: pd.DataFrame, age_start: int, age_end: int, fill_value: float):
```
kfwilson: so this checks that you don't have anything outside of the age restrictions but I don't think it also checks whether you do have all the age groups within your restrictions. are you making that check within the individual validators?

src/vivarium_inputs/validation/sim.py
```diff
@@ -142,77 +146,106 @@ def _validate_demographic_dimensions(data, entity, location):
...
+    outside = data.loc[(data.age_group_start < age_start) | (data.age_group_end > age_end)]
+    if not outside.empty:
+        if (outside.value != fill_value).any():
```
kfwilson: I think this could be collapsed into a single if block rather than nested

kfwilson: also period on the end of error messages please :)

src/vivarium_inputs/validation/sim.py
```diff
@@ -142,77 +146,106 @@ def _validate_demographic_dimensions(data, entity, location):
...
+            raise DataFormattingError(f"Age restrictions are violated by a value other than {fill_value}")
+
+
+def _check_sex_restrictions(data: pd.DataFrame, male_only: bool, female_only: bool, fill_value: float):
+    if male_only:
+        if (data.loc[data.sex == 'Female', 'value'] != fill_value).any():
```
kfwilson: I don't think these could also be collapsed instead of nested + periods on the end of these error messages too please

src/vivarium_inputs/validation/sim.py
```diff
@@ -142,77 +146,103 @@ def _validate_demographic_dimensions(data, entity, location):
...
 
-def _validate_draw_column(data):
+def _validate_draw_column(data: pd.DataFrame):
     if 'draw' not in data.columns:
-        raise DataFormattingError('Draw column name improperly specified.')
+        raise DataFormattingError("Draw data must be contained in a column named 'draw'.")
 
-    if list(data['draw'].unique()) != list(range(1000)):
-        raise DataFormattingError('Draw values improperly specified.')
+    if set(data['draw'].unique()) != set(range(1000)):
```
michaella21: you can remove unique

src/vivarium_inputs/validation/sim.py
```diff
@@ -142,77 +146,103 @@ def _validate_demographic_dimensions(data, entity, location):
...
+        raise DataFormattingError('Location must contain a single value that matches specified location.')
 
 
-def _validate_sex_column(data):
+def _validate_sex_column(data: pd.DataFrame):
     if 'sex' not in data.columns:
-        raise DataFormattingError('Sex column name improperly specified.')
+        raise DataFormattingError("Sex data must be contained in a column named 'sex'.")
 
     if set(data['sex'].unique()) != {'Male', 'Female'}:
```
michaella21: here also remove unique()?

PR #99, opened 2019-01-30
Feature/gbd 2017 core and loader

michaella21:
filter data by cause/risk restrictions (age/sex)

### DIFF
--- a/src/vivarium_inputs/core.py
+++ b/src/vivarium_inputs/core.py
@@ -288,3 +288,4 @@ def get_demographic_dimensions(entity: POP, location_id: int) -> pd.DataFrame:
     demographic_dimensions = utilities.get_demographic_dimensions(location_id)
     demographic_dimensions = utilities.normalize(demographic_dimensions)
     return demographic_dimensions
+
--- a/src/vivarium_inputs/core.py
+++ b/src/vivarium_inputs/core.py
@@ -6,7 +6,7 @@ import pandas as pd
 import numpy as np
 
 from vivarium_inputs import utilities, extract
-from vivarium_inputs.validation.utilities import get_restriction_age_boundary, get_restriction_age_ids
+#from vivarium_inputs.validation.utilities import get_restriction_age_boundary, get_restriction_age_ids
 from vivarium_inputs.mapping_extension import AlternativeRiskFactor, HealthcareEntity, HealthTechnology
 
 from .globals import InvalidQueryError, DEMOGRAPHIC_COLUMNS, DRAW_COLUMNS, MEASURES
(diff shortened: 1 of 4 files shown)


src/vivarium_inputs/validation/utilities.py
```diff
@@ -466,7 +466,7 @@ def check_metric_id(data: pd.DataFrame, expected_metric: str):
                                 f'(metric_id {METRICS[expected_metric.capitalize()]}')
 
 
-def get_restriction_age_boundary(entity: RiskFactor, boundary: str):
+def get_restriction_age_boundary(entity: RiskFactor, boundary: str, reverse=False):
```
kfwilson: should this be Union[RiskFactor, Cause] now? If so, also update doc string for entity param.

src/vivarium_inputs/validation/utilities.py
```diff
@@ -477,6 +477,9 @@ def get_restriction_age_boundary(entity: RiskFactor, boundary: str):
     boundary
         String 'start' or 'end' indicating whether to return the minimum
```
kfwilson: update this now that it's not always min start or max end

src/vivarium_inputs/validation/utilities.py
```diff
@@ -138,7 +138,7 @@ def check_data_exist(data: pd.DataFrame, zeros_missing: bool,
     return True
 
 
-def _get_restriction_ages(start_id: Union[float, None], end_id: Union[float, None]) -> list:
+def get_restriction_ages(start_id: Union[float, None], end_id: Union[float, None]) -> list:
```
kfwilson: this is my bad because I originally wrote it, but I think this should be named get_restriction_age_ids - it confused me for a bit reading it where you were using it to filter the age group id col in e.g., get_incidence 

also now that you're making this a public method, can we add a doc string since I put doc strings on all other public methods in utils?

src/vivarium_inputs/utilities.py
```diff
@@ -26,6 +28,13 @@ def get_annual_year_bins():
     return scrub_year(df)
 
 
+def get_valid_sex_ids(restrictions: Restrictions) -> List:
```
kfwilson: this means if the restrictions are male_only = False, female_only = False, you'll end up with valid sex ids of [] but that's not what you want

src/vivarium_inputs/core.py
```diff
@@ -68,6 +81,18 @@ def get_incidence(entity: Union[Cause, Sequela], location_id: int) -> pd.DataFra
 
 def get_prevalence(entity: Union[Cause, Sequela], location_id: int) -> pd.DataFrame:
     data = extract.extract_data(entity, 'prevalence', location_id)
+    if entity.kind == 'cause':
+        restrictions = entity.restrictions
+    else:  # sequela
+        cause = [c for c in causes if entity in c.sequelae][0]
+        restrictions = cause.restrictions
+
+    ages = get_restriction_ages(restrictions.yld_age_group_id_start, restrictions.yld_age_group_id_end)
```
kfwilson: maybe a function for these filters since they get used a lot?

src/vivarium_inputs/validation/utilities.py
michaella21: for now, just rename and move it out later. will add the doc strings

src/vivarium_inputs/validation/utilities.py
michaella21: yep fixed!

src/vivarium_inputs/utilities.py
```diff
@@ -245,3 +247,51 @@ def compute_categorical_paf(rr_data: pd.DataFrame, e: pd.DataFrame, affected_ent
...
+    if restrictions.male_only and not restrictions.female_only:
+        sexes = gbd.MALE
+    elif not restrictions.male_only and restrictions.female_only:
+        sexes = gbd.FEMALE
+    else:  # not male only and not female only
+        sexes = gbd.FEMALE + gbd.MALE
+
+    data = data[data.sex_id.isin(sexes)]
+
+    if which_age == 'yll':
```
kfwilson: I feel like lines 283-294 could get pulled out into a helper function that then @chorst could use rather than basically doing the same thing with in the sim validators

michaella21: done

src/vivarium_inputs/utilities.py
```diff
@@ -245,3 +247,57 @@ def compute_categorical_paf(rr_data: pd.DataFrame, e: pd.DataFrame, affected_ent
     paf['affected_entity'] = affected_entity
     paf['affected_measure'] = affected_measure
     return paf
+
+
+def get_age_group_ids_by_restriction(entity: Union[RiskFactor, Cause], which_age: str) -> (float,float):
```
kfwilson: :100:

PR #100, opened 2019-01-30
Feature/gbd 2017 sim validator pop

kfwilson:
validators for pop measures + fixing the interface get_age_bins call to not return directly from core


src/vivarium_inputs/validation/sim.py
```diff
@@ -19,10 +19,12 @@
 VALID_EXCESS_MORT_RANGE = (0.0, 120.0)  # James' head
 VALID_COST_RANGE = (0, {'healthcare_entity': 30000, 'health_technology': 50})
 VALID_UTILIZATION_RANGE = (0, 50)
+VALID_POPULATION_RANGE = (0, 100000000)
```
collijk: Write as `10_000_000`

kfwilson: done! also did the one in raw to make more readable

PR #101, opened 2019-01-30
Feature/refactor age checker

halfhorst:
I changed the age restriction checking so I could implement the more narrow checks that RR and PAF require


src/vivarium_inputs/validation/sim.py
```diff
@@ -320,15 +320,40 @@ def _validate_value_column(data: pd.DataFrame):
         raise DataFormattingError('Value data found to contain infinity.')
 
 
-def _translate_age_restrictions(ids: Sequence[int]) -> (float, float):
+def _translate_age_restrictions(start_ids: Sequence[Union[int, None]], end_ids: Sequence[Union[int, None]],
```
kfwilson: I feel like you and @michaella21 have both done pretty much the same thing. I think we should pull out the part of michelle's PR that I commented and make it a helper function that you then call to get the start/end age ids and then this function becomes just get the age_group_start and age_group_end corresponding to the those ids

halfhorst: works for me

PR #103, opened 2019-01-31
Feature/risk like sim validators

halfhorst:
Risk-like validators.

Small change to the age restriction check -- it now uses michelle's function. I had to update the calls in all the other validators, too, as well as the tests.


src/vivarium_inputs/validation/sim.py
```diff
@@ -17,6 +17,11 @@
 VALID_REMISSION_RANGE = (0.0, 120.0)  # James' head
 VALID_CAUSE_SPECIFIC_MORTALITY_RANGE = (0.0, 0.4)  # used mortality viz, picked worst country 15q45, mul by ~1.25
 VALID_EXCESS_MORT_RANGE = (0.0, 120.0)  # James' head
+VALID_EXPOSURE_RANGE = (0.0, {'continuous': 10_000.0, 'categorical': 1.0})
+VALID_EXPOSURE_SD_RANGE = (0.0, 1000.0)  # James' brain
+VALID_EXPOSURE_DIST_WEIGHTS_RANGE = (0.0, 1.0)
+VALID_RELATIVE_RISK_RANGE = (1.0, {'continuous': 5.0, 'categorical': 20.0})
```
halfhorst: I bumped up categorical. I found something that violated it almost immediately (chewing tobacco, just above 15)

kfwilson: should we bump it up in raw validators too?

src/vivarium_inputs/validation/sim.py
```diff
@@ -176,42 +171,128 @@ def _validate_excess_mortality(data: pd.DataFrame, entity: Cause, location: str)
...
+    validation_utilities.check_value_columns_boundary(data, boundary_value=VALID_EXPOSURE_RANGE[1][valid_kwd],
+                                                      boundary_type='upper', value_columns=['value'], error=True)
+
+    cats = data.groupby('parameter')
+    cats.apply(_validate_standard_columns, location)
+
+    if is_categorical:
+        non_categorical_columns = list(set(data.columns).difference({'parameter', 'value'}))
+        if not np.allclose(data.groupby(non_categorical_columns)['value'].sum(), 1.0):
+            raise DataFormattingError("Categorical exposures do not sum to one across categories")
```
kfwilson: period please

src/vivarium_inputs/validation/sim.py
```diff
@@ -176,42 +171,128 @@ def _validate_excess_mortality(data: pd.DataFrame, entity: Cause, location: str)
...
-    raise NotImplemented()
 
+def _validate_relative_risk(data: pd.DataFrame, entity: Union[RiskFactor, CoverageGap], location: str):
+    risk_relationship = data.groupby(['affected_entity', 'affected_measure', 'parameter'])
+    risk_relationship.apply(_validate_standard_columns, location)
 
-def _validate_exposure_distribution_weights(data, entity, location):
-    raise NotImplemented()
+    is_continuous = entity.distribution in ['normal', 'lognormal', 'ensemble']
+    is_categorical = (entity.distribution in ['dichotomous', 'ordered_polytomous', 'unordered_polytomous'])
```
kfwilson: same question re parens just for is_categorical but not is_continuous

src/vivarium_inputs/validation/sim.py
```diff
@@ -176,42 +171,128 @@ def _validate_excess_mortality(data: pd.DataFrame, entity: Cause, location: str)
...
 
+    if is_categorical:
+        tmrel_cat = sorted(list(entity.categories.to_dict()), key=lambda x: int(x[3:]))[-1]  # chop 'cat' and sort
+        if not np.allclose(data.loc[data.parameter == tmrel_cat, 'value'], 1.0):
+            raise DataFormattingError(f"The TMREL category {tmrel_cat} contains values other than 1.0.")
 
-def _validate_population_attributable_fraction(data, entity, location):
-    raise NotImplemented()
+    if data.affected_measure == 'incidence_rate':
+        _check_age_restrictions(data, entity, type='inner', fill_value=1.0)
```
kfwilson: why inner and not yld?

src/vivarium_inputs/validation/sim.py
```diff
@@ -176,42 +171,128 @@ def _validate_excess_mortality(data: pd.DataFrame, entity: Cause, location: str)
...
+def _validate_population_attributable_fraction(data: pd.DataFrame, entity: Union[RiskFactor, Etiology], location: str):
+    risk_relationship = data.groupby(['affected_entity', 'affected_measure'])
+    risk_relationship.apply(_validate_standard_columns, location)
+
+    validation_utilities.check_value_columns_boundary(data, boundary_value=VALID_PAF_RANGE[0],
+                                                      boundary_type='lower', value_columns=['value'], error=True)
+    validation_utilities.check_value_columns_boundary(data, boundary_value=VALID_PAF_RANGE[1],
+                                                      boundary_type='upper', value_columns=['value'], error=True)
+
+    _check_age_restrictions(data, entity, type='inner', fill_value=0.0)
```
kfwilson: I think from what @michaella21 found in digging into paf data for child_stunting we actually should be grouping by affected entity and checking the restrictions of the affected entity. But also, we're getting restrictions for measures in three different places now (raw validators, core to filter, and sim validators) and I'm not confident it's all lined up. Probably would be worth pulling that out into a util method that we can use in all three places - that's probably not something to do in this PR but as a separate one

halfhorst: This is the sort of thing that seems like should be done in a pass once we are all done. It's difficult to conceptualize the right way to abstract things with a moving target

halfhorst: This seems like the sort of thing that should be done in one pass after. It's difficult to figure out the right abstractions with a moving target

halfhorst: nice, a double comment :man_facepalming:

src/vivarium_inputs/validation/sim.py
halfhorst: we just decided to match RR

src/vivarium_inputs/validation/sim.py
halfhorst: I'd vote ya unless we have a compelling reason to worry about values in between 15 and 20

kfwilson: I agree. Can you add that in this PR? It's just changing the constant at the top of raw.py - I don't want to open a separate PR just for that

src/vivarium_inputs/validation/sim.py
```diff
@@ -105,9 +108,7 @@ def _validate_prevalence(data: pd.DataFrame, entity: Union[Cause, Sequela], loca
     validation_utilities.check_value_columns_boundary(data, boundary_value=VALID_PREVALENCE_RANGE[1],
                                                       boundary_type='upper', value_columns=['value'], error=True)
 
-    age_start, age_end = _translate_age_restrictions((entity.restrictions.yld_age_group_id_start,
-                                                      entity.restrictions.yld_age_group_id_end))
-    _check_age_restrictions(data, age_start, age_end, fill_value=0.0)
+    _check_age_restrictions(data, entity, type='yld', fill_value=0.0)
```
kfwilson: this is rest_type in your function def

src/vivarium_inputs/validation/sim.py
```diff
@@ -176,42 +171,128 @@ def _validate_excess_mortality(data: pd.DataFrame, entity: Cause, location: str)
...
-def _validate_relative_risk(data, entity, location):
-    raise NotImplemented()
+    validation_utilities.check_value_columns_boundary(data, boundary_value=VALID_RELATIVE_RISK_RANGE[0],
+                                                      boundary_type='lower', value_columns=['value'], error=True)
+    validation_utilities.check_value_columns_boundary(data, boundary_value=VALID_RELATIVE_RISK_RANGE[1][range_kwd],
+                                                      boundary_type='upper', value_columns=['value'], error=True)
 
+    if is_categorical:
+        tmrel_cat = sorted(list(entity.categories.to_dict()), key=lambda x: int(x[3:]))[-1]  # chop 'cat' and sort
+        if not np.allclose(data.loc[data.parameter == tmrel_cat, 'value'], 1.0):
```
collijk: Should be exactly 1.  Do a is close check in core an set if meets tolerance.

PR #107, opened 2019-01-31
Feature/test update utilities

halfhorst:
This is an orphaned branch. I started writing some tests for utilities and then stopped and did the sim validators. But, this found a bug, so may as well pull it in rather than discarding it.

Also, I changed a place in utilities to use a call to vivarium gbd access instead of a hard-coded year list.


src/vivarium_inputs/utilities.py
```diff
@@ -137,7 +137,8 @@ def normalize_sex(data: pd.DataFrame, fill_value) -> pd.DataFrame:
 
 
 def normalize_year(data: pd.DataFrame) -> pd.DataFrame:
-    years = {'annual': list(range(1990, 2018)), 'binned': gbd.get_estimation_years()}
+    binned_years = gbd.get_estimation_years()
```
halfhorst: Ideally this would be linked up with get_annual_year_bins so get_estimation_years is only ever called in one place and we have functions that define our estimation periods. But that feels like a passthrough-after kind of thing.

PR #109, opened 2019-01-31
Bugfix/restriction filtering

halfhorst:
A few more bugfixes

### DIFF
--- a/src/vivarium_inputs/utilities.py
+++ b/src/vivarium_inputs/utilities.py
@@ -138,7 +138,7 @@ def normalize_sex(data: pd.DataFrame, fill_value) -> pd.DataFrame:
 
 def normalize_year(data: pd.DataFrame) -> pd.DataFrame:
     binned_years = gbd.get_estimation_years()
-    years = {'annual': list(range(binned_years.min(), binned_years.max() + 1)), 'binned': binned_years}
+    years = {'annual': list(range(min(binned_years), max(binned_years) + 1)), 'binned': binned_years}
 
     if 'year_id' not in data:
         # Data doesn't vary by year, so copy for each year.
--- a/src/vivarium_inputs/utilities.py
+++ b/src/vivarium_inputs/utilities.py
@@ -294,7 +294,7 @@ def filter_data_by_restrictions(data: pd.DataFrame, entity: Union[RiskFactor, Ca
     elif not restrictions.male_only and restrictions.female_only:
         sexes = gbd.FEMALE
     else:  # not male only and not female only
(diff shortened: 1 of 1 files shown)


src/vivarium_inputs/utilities.py
```diff
@@ -138,7 +138,7 @@ def normalize_sex(data: pd.DataFrame, fill_value) -> pd.DataFrame:
 
 def normalize_year(data: pd.DataFrame) -> pd.DataFrame:
     binned_years = gbd.get_estimation_years()
-    years = {'annual': list(range(binned_years.min(), binned_years.max() + 1)), 'binned': binned_years}
+    years = {'annual': list(range(min(binned_years), max(binned_years) + 1)), 'binned': binned_years}
```
kfwilson: why is this the right min/max to use? I didn't realize they would give different results

src/vivarium_inputs/utilities.py
```diff
@@ -294,7 +294,7 @@ def filter_data_by_restrictions(data: pd.DataFrame, entity: Union[RiskFactor, Ca
     elif not restrictions.male_only and restrictions.female_only:
         sexes = gbd.FEMALE
     else:  # not male only and not female only
-        sexes = gbd.FEMALE + gbd.MALE
+        sexes = gbd.FEMALE + gbd.MALE + gbd.COMBINED
```
kfwilson: hmm wait actually where are we using this that we have sex_id = 3 and why do we need to keep those rows?

src/vivarium_inputs/utilities.py
michaella21: I included in current my working branch also :) yeah we can't use the .min() for the list :(

src/vivarium_inputs/utilities.py
michaella21: I think it's for the case when data frame only has sex id 3?

src/vivarium_inputs/utilities.py
halfhorst: it is a plain python list, not a series or dataframe

kfwilson: ohh didn't realize get_estimation_years gave back a list

src/vivarium_inputs/utilities.py
halfhorst: this filtering happens before normalization, so sex_id 3 still can exist in the data

halfhorst: it led to an empty dataframe. Incidentally, maybe we should catch that -- if you filter everything away, raise some sort of exception

PR #113, opened 2019-02-01
Bugfix/rr flags checker

halfhorst:
This handles the case when a cause id in the data does not match anything in our map


collijk:
Hold til we see if switching hierarchies fixes issues

src/vivarium_inputs/validation/raw.py
```diff
@@ -490,8 +490,12 @@ def _validate_relative_risk(data: pd.DataFrame, entity: Union[RiskFactor, Covera
     check_value_columns_boundary(data, max_val, 'upper', value_columns=DRAW_COLUMNS, inclusive=True, error=None)
 
     for c_id in data.cause_id.unique():
```
collijk:
```
for cause_id in set(data.cause_id).intersection([c.gbd_id for c in causes]):
    check_mort_morb_flags(...)
```

collijk:
Michelle is handling in a separate PR

PR #114, opened 2019-02-01
Bugfix/raw validator measure dict

halfhorst:
A small bug that turned up. We were comparing an int to string


src/vivarium_inputs/validation/raw.py
```diff
@@ -529,10 +529,10 @@ def _validate_population_attributable_fraction(data: pd.DataFrame, entity: Union
 
     for c_id in data.cause_id:
         cause = [c for c in causes if c.gbd_id == c_id][0]
-        if cause.restrictions.yld_only and (data.measure_id == 'YLLs').any():
+        if cause.restrictions.yld_only and (data.measure_id == MEASURES['YLLs']).any():
```
kfwilson: oh yeah this is my bad. I think it's also in Michelle's PR though

PR #122, opened 2019-02-04
Feature/gbd 2017 rr update

michaella21:
update for RR, all the check in validator only using exposure data

### DIFF
--- a/src/vivarium_inputs/core.py
+++ b/src/vivarium_inputs/core.py
@@ -1,5 +1,6 @@
 from collections import namedtuple
 from typing import Union
+from itertools import product
 
 from gbd_mapping import Cause, Sequela, RiskFactor, CoverageGap, Etiology, Covariate, causes
 import pandas as pd
@@ -203,18 +204,24 @@ def get_exposure_distribution_weights(entity: Union[RiskFactor, AlternativeRiskF
     return data
 
 
-def _handle_relative_risk_age_groups(data):
-    causes_map = {c.gbd_id: c for c in causes}
+def filter_relative_risk_to_cause_restrictions(data: pd.DataFrame) -> pd.DataFrame:
+    """ It applies age restrictions according to affected causes
+    and affected measures. If affected measure is incidence_rate,
+    it applies the yld_age_restrictions. If affected measure is
+    excess_mortality, it applies the yll_age_restrictions to filter
+    the relative_risk data"""
+
+    causes_map = {c.name: c for c in causes}
     temp = []
-    for c_id in set(data.cause_id):
-        df = data[data.cause_id == c_id]
-        cause = causes_map[c_id]
-        if cause.restrictions.yll_only:
+    affected_entities = set(data.affected_entity)
+    affected_measures = set(data.affected_measure)
+    for cause, measure in product(affected_entities, affected_measures):
+        df = data[(data.affected_entity == cause) & (data.affected_measure)]
+        cause = causes_map[cause]
+        if measure == 'excess_mortality':
             start, end = utilities.get_age_group_ids_by_restriction(cause, 'yll')
-        elif cause.restrictions.yld_only:
+        else:  # incidence_rate
             start, end = utilities.get_age_group_ids_by_restriction(cause, 'yld')
-        else:
-            start, end = utilities.get_age_group_ids_by_restriction(cause, 'inner')
         temp.append(df[df.age_group_id.isin(range(start, end + 1))])
     data = pd.concat(temp)
     return data
@@ -224,13 +231,13 @@ def get_relative_risk(entity: Union[RiskFactor, CoverageGap], location_id: int) 
     data = extract.extract_data(entity, 'relative_risk', location_id)
 
     if entity.kind == 'risk_factor':
-        data = _handle_relative_risk_age_groups(data)
         data = utilities.convert_affected_entity(data, 'cause_id')
         morbidity = data.morbidity == 1
         mortality = data.mortality == 1
         data.loc[morbidity & mortality, 'affected_measure'] = 'incidence_rate'
         data.loc[morbidity & ~mortality, 'affected_measure'] = 'incidence_rate'
         data.loc[~morbidity & mortality, 'affected_measure'] = 'excess_mortality'
+        data = filter_relative_risk_to_cause_restrictions(data)
     else:  # coverage_gap
         data = utilities.convert_affected_entity(data, 'rei_id')
         data['affected_measure'] = 'exposure_parameters'
--- a/src/vivarium_inputs/core.py
+++ b/src/vivarium_inputs/core.py
@@ -199,6 +199,9 @@ def get_exposure_distribution_weights(entity: Union[RiskFactor, AlternativeRiskF
 
 def get_relative_risk(entity: Union[RiskFactor, CoverageGap], location_id: int) -> pd.DataFrame:
     data = extract.extract_data(entity, 'relative_risk', location_id)
+    data = utilities.filter_to_most_detailed_causes(data)
+    exposure = extract.extract_data(entity, 'exposure', location_id)
+
     if entity.kind == 'risk_factor':
         data = utilities.filter_data_by_restrictions(data, entity, 'inner')
 
--- a/src/vivarium_inputs/core.py
+++ b/src/vivarium_inputs/core.py
@@ -200,11 +200,8 @@ def get_exposure_distribution_weights(entity: Union[RiskFactor, AlternativeRiskF
 def get_relative_risk(entity: Union[RiskFactor, CoverageGap], location_id: int) -> pd.DataFrame:
     data = extract.extract_data(entity, 'relative_risk', location_id)
     data = utilities.filter_to_most_detailed_causes(data)
-    exposure = extract.extract_data(entity, 'exposure', location_id)
 
     if entity.kind == 'risk_factor':
-        data = utilities.filter_data_by_restrictions(data, entity, 'inner')
-
         data = utilities.convert_affected_entity(data, 'cause_id')
         morbidity = data.morbidity == 1
         mortality = data.mortality == 1
--- a/src/vivarium_inputs/core.py
+++ b/src/vivarium_inputs/core.py
@@ -205,11 +205,29 @@ def get_exposure_distribution_weights(entity: Union[RiskFactor, AlternativeRiskF
     return data
 
 
+def _handle_relative_risk_age_groups(data):
+    causes_map = {c.gbd_id: c for c in causes}
+    temp = []
+    for c_id in set(data.cause_id):
+        df = data[data.cause_id == c_id]
+        cause = causes_map[c_id]
+        if cause.restrictions.yll_only:
+            start, end = utilities.get_age_group_ids_by_restriction(cause, 'yll')
+        elif cause.restrictions.yld_only:
+            start, end = utilities.get_age_group_ids_by_restriction(cause, 'yld')
+        else:
+            start, end = utilities.get_age_group_ids_by_restriction(cause, 'inner')
+        temp.append(df[df.age_group_id.isin(range(start, end + 1))])
+    data = pd.concat(temp)
+    return data
+
+
 def get_relative_risk(entity: Union[RiskFactor, CoverageGap], location_id: int) -> pd.DataFrame:
     data = extract.extract_data(entity, 'relative_risk', location_id)
     data = utilities.filter_to_most_detailed_causes(data)
 
     if entity.kind == 'risk_factor':
+        data = _handle_relative_risk_age_groups(data)
         data = utilities.convert_affected_entity(data, 'cause_id')
         morbidity = data.morbidity == 1
         mortality = data.mortality == 1
--- a/src/vivarium_inputs/core.py
+++ b/src/vivarium_inputs/core.py
@@ -222,7 +222,6 @@ def _handle_relative_risk_age_groups(data):
 
 def get_relative_risk(entity: Union[RiskFactor, CoverageGap], location_id: int) -> pd.DataFrame:
     data = extract.extract_data(entity, 'relative_risk', location_id)
-    data = utilities.filter_to_most_detailed_causes(data)
 
     if entity.kind == 'risk_factor':
         data = _handle_relative_risk_age_groups(data)
--- a/src/vivarium_inputs/validation/raw.py
+++ b/src/vivarium_inputs/validation/raw.py
@@ -564,8 +564,9 @@ def validate_relative_risk(data: pd.DataFrame, entity: Union[RiskFactor, Coverag
         grouped.apply(check_age_group_ids, age_start, age_end)
         grouped.apply(check_sex_ids, male_expected, female_expected)
 
-        #  We cannot check age_restrictions with exposure_age_groups since RR may have a subset of age_group_ids and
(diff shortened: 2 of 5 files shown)


src/vivarium_inputs/validation/raw.py
```diff
@@ -450,8 +452,14 @@ def _validate_exposure_distribution_weights(data: pd.DataFrame, entity: Union[Ri
...
 
 
-def _validate_relative_risk(data: pd.DataFrame, entity: Union[RiskFactor, CoverageGap], location_id: int):
-    check_data_exist(data, zeros_missing=True)
+def _validate_relative_risk(data: pd.DataFrame, entity: Union[RiskFactor, CoverageGap], location_id: int,
+                            exposure: pd.DataFrame):
+
+    data = filter_to_most_detailed_causes(data)
+    exposure_age_groups = set(exposure.age_group_id)
+    data.groupby(['cause_id', 'morbidity']).apply(lambda df: _check_age_groups_relative_risk(df, exposure_age_groups))
```
michaella21: I did this as the first check, so that we know once passed, all the age groups in data are valid age groups (subset of exposure age_groups)

kfwilson: hmm ok so before we were checking age restrictions (and age groups) by category. I think we still want to do that (i.e., add 'parameter' to your groupby clause)

src/vivarium_inputs/validation/raw.py
```diff
@@ -466,25 +474,24 @@ def _validate_relative_risk(data: pd.DataFrame, entity: Union[RiskFactor, Covera
...
     if entity.kind == 'risk_factor':
         restrictions = entity.restrictions
-        age_start = get_restriction_age_boundary(entity, 'start')
-        age_end = get_restriction_age_boundary(entity, 'end')
+        age_start = min(exposure_age_groups)
+        age_end = max(exposure_age_groups)
         male_expected = restrictions.male_only or (not restrictions.male_only and not restrictions.female_only)
         female_expected = restrictions.female_only or (not restrictions.male_only and not restrictions.female_only)
 
         cats.apply(check_age_group_ids, age_start, age_end)
```
kfwilson: I'm a little confused about why we're now checking some things by cause and morbidity and other things just by parameter. It seems like we should always be using the same groupby columns

src/vivarium_inputs/validation/raw.py
kfwilson: But we're still doing the check_age_group_ids by parameter which will also check if the age groups are valid. i don't think this has to happen in a different order than in all the other validators. I also think we still want to be checking the other standards (e.g., the data exist, the columns are what we expect) before we start doing column specific checks like this

src/vivarium_inputs/core.py
```diff
@@ -199,9 +199,9 @@ def get_exposure_distribution_weights(entity: Union[RiskFactor, AlternativeRiskF
 
 def get_relative_risk(entity: Union[RiskFactor, CoverageGap], location_id: int) -> pd.DataFrame:
     data = extract.extract_data(entity, 'relative_risk', location_id)
-    if entity.kind == 'risk_factor':
-        data = utilities.filter_data_by_restrictions(data, entity, 'inner')
+    data = utilities.filter_to_most_detailed_causes(data)
```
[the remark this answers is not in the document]
collijk: In particular, the validator shouldn't be filtering data.  That breaks our abstractions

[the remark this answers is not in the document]
michaella21: this is core, validator does use filtered data, but it doesn't really return filtered data

src/vivarium_inputs/validation/raw.py
```diff
@@ -466,25 +474,24 @@ def _validate_relative_risk(data: pd.DataFrame, entity: Union[RiskFactor, Covera
 
     if entity.kind == 'risk_factor':
         restrictions = entity.restrictions
-        age_start = get_restriction_age_boundary(entity, 'start')
-        age_end = get_restriction_age_boundary(entity, 'end')
+        age_start = min(exposure_age_groups)
+        age_end = max(exposure_age_groups)
         male_expected = restrictions.male_only or (not restrictions.male_only and not restrictions.female_only)
```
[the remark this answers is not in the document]
kfwilson: I wrote it like this because if both male_only and female_only are False, we want male_expected and female_expected to be True

src/vivarium_inputs/core.py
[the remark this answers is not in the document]
collijk: Sorry, misplaced this comment.  You're filtering data in the raw validator.

src/vivarium_inputs/validation/raw.py
```diff
@@ -450,8 +452,14 @@ def _validate_exposure_distribution_weights(data: pd.DataFrame, entity: Union[Ri
         raise DataAbnormalError(f'Distribution weights for {entity.kind} {entity.name} do not sum to 1.')
 
 
-def _validate_relative_risk(data: pd.DataFrame, entity: Union[RiskFactor, CoverageGap], location_id: int):
-    check_data_exist(data, zeros_missing=True)
+def _validate_relative_risk(data: pd.DataFrame, entity: Union[RiskFactor, CoverageGap], location_id: int,
+                            exposure: pd.DataFrame):
+
+    data = filter_to_most_detailed_causes(data)
```
collijk: This is the filter I mean.

src/vivarium_inputs/core.py
```diff
@@ -205,11 +205,29 @@ def get_exposure_distribution_weights(entity: Union[RiskFactor, AlternativeRiskF
...
+
+
 def get_relative_risk(entity: Union[RiskFactor, CoverageGap], location_id: int) -> pd.DataFrame:
     data = extract.extract_data(entity, 'relative_risk', location_id)
-    if entity.kind == 'risk_factor':
-        data = utilities.filter_data_by_restrictions(data, entity, 'inner')
+    data = utilities.filter_to_most_detailed_causes(data)
 
+    if entity.kind == 'risk_factor':
+        data = _handle_relative_risk_age_groups(data)
```
michaella21: this filtering was missing previously

src/vivarium_inputs/validation/raw.py
michaella21: yes, we just should check all 4 standards including mortality/morbidity and parameter

src/vivarium_inputs/validation/raw.py
michaella21: no, actually @kfwilson is correct, I first thought that it will raise errors for the case RR has a smaller age group id set than exposure from the check_age_group_ids, but it isn't and we can just use that function to check.

src/vivarium_inputs/core.py
```diff
@@ -203,11 +203,29 @@ def get_exposure_distribution_weights(entity: Union[RiskFactor, AlternativeRiskF
...
+        temp.append(df[df.age_group_id.isin(range(start, end + 1))])
+    data = pd.concat(temp)
+    return data
+
+
 def get_relative_risk(entity: Union[RiskFactor, CoverageGap], location_id: int) -> pd.DataFrame:
     data = extract.extract_data(entity, 'relative_risk', location_id)
-    if entity.kind == 'risk_factor':
-        data = utilities.filter_data_by_restrictions(data, entity, 'inner')
+    data = utilities.filter_to_most_detailed_causes(data)
```
kfwilson: wait aren't you already doing this filtering in extract?

src/vivarium_inputs/validation/raw.py
```diff
@@ -450,9 +452,17 @@ def _validate_exposure_distribution_weights(data: pd.DataFrame, entity: Union[Ri
...
 
+    for c_id in data.cause_id.unique():
+        cause = [c for c in causes if c.gbd_id == c_id][0]
+        check_mort_morb_flags(data, cause.restrictions.yld_only, cause.restrictions.yll_only)
+
+    exposure_age_groups = set(exposure.age_group_id)
+
     expected_columns = ['rei_id', 'modelable_entity_id', 'cause_id', 'mortality',
                         'morbidity', 'metric_id', 'parameter'] + DEMOGRAPHIC_COLUMNS + DRAW_COLUMNS
     check_columns(expected_columns, data.columns)
```
kfwilson: I still think we want the expected columns check to come before the mort/morb checks. The columns check should be right after the data exist check because if we're missing say mortality or morbidity, we're going to get a much more confusing error if we jump right into check_mort_morb_flags

src/vivarium_inputs/validation/raw.py
```diff
@@ -462,33 +472,27 @@ def _validate_relative_risk(data: pd.DataFrame, entity: Union[RiskFactor, Covera
...
         female_expected = restrictions.female_only or (not restrictions.male_only and not restrictions.female_only)
 
-        cats.apply(check_age_group_ids, age_start, age_end)
-        cats.apply(check_sex_ids, male_expected, female_expected)
+        grouped.apply(check_age_group_ids, age_start, age_end)
+        grouped.apply(check_sex_ids, male_expected, female_expected)
 
-        cats.apply(check_age_restrictions, age_start, age_end)
-        cats.apply(check_sex_restrictions, entity.restrictions.male_only, entity.restrictions.female_only)
+        grouped.apply(check_sex_restrictions, entity.restrictions.male_only, entity.restrictions.female_only)
```
kfwilson: hmm don't we still want to do check_age_restrictions with the exposure age groups? check_age_group_ids just checks the age group id column; check_age_restrictions also checks the draw columns

src/vivarium_inputs/validation/sim.py
```diff
@@ -280,8 +280,7 @@ def _validate_relative_risk(data: pd.DataFrame, entity: Union[RiskFactor, Covera
         raise NotImplementedError()
 
     validation_utilities.check_value_columns_boundary(data, boundary_value=VALID_RELATIVE_RISK_RANGE[0],
-                                                      boundary_type='lower', value_columns=['value'],
-                                                      error=DataTransformationError)
+                                                      boundary_type='lower', value_columns=['value'])
```
kfwilson: Wait I think we still want to throw an error here in sim validators? If RR isn't meeting the lower boundary in sim validators, do we really just want to warn and carry on?

michaella21: on Thursday coworking, we decided to give warning, (I don't know whether we want to do it temporarily or forever thou) We probably revisit it since some risk has low boundary a little off from 1 while the other is very off from 1

src/vivarium_inputs/validation/raw.py
michaella21: ok, then I think this mort/morb flags checks can happen right before grouping

src/vivarium_inputs/validation/sim.py
collijk: I think we want a soft rr boundary at 1 (some risks legitimately have protective effects) and a hard rr boundary at 0 (non-inclusive) since RR <=0 doesn't make sense.

src/vivarium_inputs/validation/raw.py
michaella21: I'd like to. But RR can have a smaller set of age group_ids then exposure, and we can't raise an error but check_age_restrictions will raise an error for the missing age groups if we use exposure age groups. The main check we do in RR is whether RR has a bigger set of age group ids than exposure.

kfwilson: oh ok got it. Can we add a comment about this then? I can definitely seeing myself forgetting this in a few months and trying to add it back in because I thought it was forgotten

michaella21: yep, will add it

src/vivarium_inputs/validation/raw.py
collijk: Agree.  Maybe
- check data
- check columns
- check metric
- check years
- check location
- group
- check sex column
- check age column
- check sex restrictions
- check age restrictions
- check values

src/vivarium_inputs/core.py
```diff
@@ -203,11 +203,28 @@ def get_exposure_distribution_weights(entity: Union[RiskFactor, AlternativeRiskF
     return data
 
 
+def _handle_relative_risk_age_groups(data):
```
collijk: Let's name this better.  filter_relative_risk_to_cause_restrictions?  typing and a docstring.

src/vivarium_inputs/core.py
```diff
@@ -203,11 +203,28 @@ def get_exposure_distribution_weights(entity: Union[RiskFactor, AlternativeRiskF
...
+    temp = []
+    for c_id in set(data.cause_id):
+        df = data[data.cause_id == c_id]
+        cause = causes_map[c_id]
+        if cause.restrictions.yll_only:
+            start, end = utilities.get_age_group_ids_by_restriction(cause, 'yll')
+        elif cause.restrictions.yld_only:
+            start, end = utilities.get_age_group_ids_by_restriction(cause, 'yld')
+        else:
+            start, end = utilities.get_age_group_ids_by_restriction(cause, 'inner')
```
collijk: I think we have multiple cases here.  rr data with the morbidity flag should be filtered to the yld restrictions.  rr data with the mortality flag should be filtered to the yll restrictions.  A comment explaining what's happening would probably be helpful reading this later too.

collijk: Maybe do the filter after you split the data to make things easier.

src/vivarium_inputs/validation/raw.py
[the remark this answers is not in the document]
collijk: I don't understand how that would break the above code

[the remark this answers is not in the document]
collijk: @kfwilson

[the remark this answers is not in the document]
kfwilson: It wouldn't. I didn't read your comment carefully enough. sorry!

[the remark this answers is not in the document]
kfwilson: I used this pattern in a lot of validators though so I'll need to change it in most of the other validators.

src/vivarium_inputs/validation/sim.py
```diff
@@ -279,8 +279,11 @@ def _validate_relative_risk(data: pd.DataFrame, entity: Union[RiskFactor, Covera
     else:
         raise NotImplementedError()
 
+    #  We want to have hard lower limit 0 for RR and soft limit 1 for RR
```
collijk: Add that the soft lower limit is because some risks are protective against some causes

src/vivarium_inputs/validation/raw.py
```diff
@@ -543,33 +549,33 @@ def validate_relative_risk(data: pd.DataFrame, entity: Union[RiskFactor, Coverag
...
 
-        cats.apply(check_age_group_ids, age_start, age_end)
-        cats.apply(check_sex_ids, male_expected, female_expected)
+        grouped.apply(check_age_group_ids, age_start, age_end)
+        grouped.apply(check_sex_ids, male_expected, female_expected)
 
-        cats.apply(check_age_restrictions, age_start, age_end)
-        cats.apply(check_sex_restrictions, entity.restrictions.male_only, entity.restrictions.female_only)
+        #  We cannot check age_restrictions with exposure_age_groups since RR may have a subset of age_group_ids and
+        #  we do not want to raise an error for this case.
```
collijk: Add why we don't want to raise the error

PR #126, opened 2019-02-04
sequela with no dw data return zeroes

halfhorst:
If metadata tells us a sequela has no disability weight associated with it, assign it zeroes.

### DIFF
--- a/src/vivarium_inputs/core.py
+++ b/src/vivarium_inputs/core.py
@@ -110,7 +110,9 @@ def get_disability_weight(entity: Union[Cause, Sequela], location_id: int) -> pd
         data = data.reset_index()
     else:  # entity.kind == 'sequela'
         if not entity.healthstate.disability_weight_exists:
-            cols = {'location_id': [location_id], 'sex_id': [3], 'age_group_id': [22]}
+            all_ages_id = 22
+            both_sexes_id = 3
+            cols = {'location_id': [location_id], 'sex_id': [both_sexes_id], 'age_group_id': [all_ages_id]}
             cols.update({f'draw_{i}': [0.0] for i in range(1000)})
             data = pd.DataFrame(cols)
         else:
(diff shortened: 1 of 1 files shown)


src/vivarium_inputs/core.py
```diff
@@ -109,7 +109,12 @@ def get_disability_weight(entity: Union[Cause, Sequela], location_id: int) -> pd
                 data += prevalence * disability
         data = data.reset_index()
     else:  # entity.kind == 'sequela'
-        data = extract.extract_data(entity, 'disability_weight', location_id)
+        if not entity.healthstate.disability_weight_exists:
+            cols = {'location_id': [location_id], 'sex_id': [3], 'age_group_id': [22]}
```
kfwilson: can we use constants or at least named vars for 3 and 22?

halfhorst: used  named vars unless we decide we want to introduce these as globals

src/vivarium_inputs/core.py
```diff
@@ -109,7 +109,14 @@ def get_disability_weight(entity: Union[Cause, Sequela], location_id: int) -> pd
                 data += prevalence * disability
         data = data.reset_index()
     else:  # entity.kind == 'sequela'
-        data = extract.extract_data(entity, 'disability_weight', location_id)
+        if not entity.healthstate.disability_weight_exists:
+            all_ages_id = 22
```
collijk: Put these constants in `globals` (also grep for 22 and 3 and replace them with the constant).

collijk: You could also use `utilities.get_demographic_dimensions` to get the full demography, and append a len(demography) x len(draws) block of zeroes to the returned frame.

src/vivarium_inputs/core.py
```diff
@@ -108,7 +108,12 @@ def get_disability_weight(entity: Union[Cause, Sequela], location_id: int) -> pd
                 data += prevalence * disability
         data = data.reset_index()
     else:  # entity.kind == 'sequela'
-        data = extract.extract_data(entity, 'disability_weight', location_id)
+        if not entity.healthstate.disability_weight_exists:
+            full_demog = utilities.get_demographic_dimensions(location_id)
+            draws = pd.DataFrame({f'draw_{i}': len(full_demog) * [0.0] for i in range(1000)})
```
collijk: Use the ``DRAW_COLUMNS`` global for the column names.

PR #128, opened 2019-02-04
Cleaning up dependencies on external data.  Ages and sexes

collijk:
Tests are broken because we've introduced a dependence on the `gbd` module.  I'm working through cleaning up dependencies generally, but the pr was starting to get hairy, so here is the first piece.


halfhorst:
thanks for doing this, was putting off the disability weight edits to pull out the constants

collijk:
@kfwilson I'll bring in those fixes with the next pr

PR #131, opened 2019-02-04
Feature/gbd 2017 paf update

michaella21:
paf update to check its age groups with RR age groups

### DIFF
--- a/src/vivarium_inputs/core.py
+++ b/src/vivarium_inputs/core.py
@@ -263,8 +263,7 @@ def get_population_attributable_fraction(entity: Union[RiskFactor, Etiology], lo
 
     if entity.kind == 'etiology':
         cause = [c for c in causes if c.etiologies and entity in c.etiologies][0]
-        restriction_entity = cause
-        data = utilities.filter_data_by_restrictions(data, restriction_entity, 'inner')
+        data = utilities.filter_data_by_restrictions(data, cause, 'inner')
 
     data = utilities.convert_affected_entity(data, 'cause_id')
     data.loc[data['measure_id'] == MEASURES['YLLs'], 'affected_measure'] = 'excess_mortality'
--- a/src/vivarium_inputs/core.py
+++ b/src/vivarium_inputs/core.py
@@ -269,7 +269,7 @@ def get_population_attributable_fraction(entity: Union[RiskFactor, Etiology], lo
     data = utilities.convert_affected_entity(data, 'cause_id')
     data.loc[data['measure_id'] == MEASURES['YLLs'], 'affected_measure'] = 'excess_mortality'
     data.loc[data['measure_id'] == MEASURES['YLDs'], 'affected_measure'] = 'incidence_rate'
-    data = data.groupby('measure_id').apply(lambda df: utilities.normalize(df, fill_value=0))
+    data = data.groupby(['affected_entity', 'affected_measure']).apply(lambda df: utilities.normalize(df, fill_value=0))
     data = utilities.reshape(data, to_keep=DEMOGRAPHIC_COLUMNS + ['affected_entity', 'affected_measure'])
     return data
 
--- a/src/vivarium_inputs/core.py
+++ b/src/vivarium_inputs/core.py
@@ -162,8 +162,8 @@ def get_exposure(entity: Union[RiskFactor, AlternativeRiskFactor, CoverageGap], 
 
     if entity.distribution in ['dichotomous', 'ordered_polytomous', 'unordered_polytomous']:
         tmrel_cat = sorted(list(entity.categories.to_dict()), key=lambda x: int(x[3:]))[-1]
-        exposed = data[~data.isin(tmrel_cat)]
-        unexposed = data[data.isin(tmrel_cat)]
+        exposed = data[~data.isin([tmrel_cat])]
+        unexposed = data[data.isin([tmrel_cat])]
         #  FIXME: We fill 1 as exposure of tmrel category, which is not correct.
         data = pd.concat([utilities.normalize(exposed, fill_value=0), utilities.normalize(unexposed, fill_value=1)],
                          ignore_index=True)
--- a/src/vivarium_inputs/core.py
+++ b/src/vivarium_inputs/core.py
@@ -261,13 +261,10 @@ def get_relative_risk(entity: Union[RiskFactor, CoverageGap], location_id: int) 
 def get_population_attributable_fraction(entity: Union[RiskFactor, Etiology], location_id: int) -> pd.DataFrame:
     data = extract.extract_data(entity, 'population_attributable_fraction', location_id)
 
-    if entity.kind == 'risk_factor':
-        restriction_entity = entity
-    else:  # etiology
+    if entity.kind == 'etiology':
         cause = [c for c in causes if c.etiologies and entity in c.etiologies][0]
         restriction_entity = cause
-
-    data = utilities.filter_data_by_restrictions(data, restriction_entity, 'inner')
+        data = utilities.filter_data_by_restrictions(data, restriction_entity, 'inner')
 
     data = utilities.convert_affected_entity(data, 'cause_id')
     data.loc[data['measure_id'] == MEASURES['YLLs'], 'affected_measure'] = 'excess_mortality'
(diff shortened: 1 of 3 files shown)


src/vivarium_inputs/core.py
```diff
@@ -261,18 +261,15 @@ def get_relative_risk(entity: Union[RiskFactor, CoverageGap], location_id: int)
 def get_population_attributable_fraction(entity: Union[RiskFactor, Etiology], location_id: int) -> pd.DataFrame:
     data = extract.extract_data(entity, 'population_attributable_fraction', location_id)
 
-    if entity.kind == 'risk_factor':
-        restriction_entity = entity
-    else:  # etiology
+    if entity.kind == 'etiology':
         cause = [c for c in causes if c.etiologies and entity in c.etiologies][0]
         restriction_entity = cause
```
halfhorst: do you need this var?

src/vivarium_inputs/validation/raw.py
```diff
@@ -592,15 +592,19 @@ def validate_population_attributable_fraction(data: pd.DataFrame, entity: Union[
...
+    age_start = get_restriction_age_boundary(restrictions_entity, 'start') if entity.kind == 'etiology' else None
+    age_end = get_restriction_age_boundary(restrictions_entity, 'end') if entity.kind == 'etiology' else None
+    male_expected = not restrictions.female_only
+    female_expected = not restrictions.male_only
 
     check_age_group_ids(data, age_start, age_end)
     check_sex_ids(data, male_expected, female_expected)
 
-    check_age_restrictions(data, age_start, age_end)
+    # We cannot check risk_factor paf age restrictions using RR age groups here,
```
collijk: Isn't this exactly what you're doing with the groupby?

src/vivarium_inputs/validation/raw.py
```diff
@@ -862,3 +867,25 @@ def check_cause_age_restrictions(entity: Cause) -> None:
...
+    """Raise an error if relative risk includes any age_group_ids with non
+    trivial data. But paf still can have an extra data while there is no
+    relative risk matching with.
+    """
+    measure_map = {MEASURES['YLLs']: 'YLLs', MEASURES['YLDs']: 'YLDs'}
+    cid = paf.cause_id.unique()[0]
+    measure = paf.measure_id.unique()[0]
+
+    if measure == MEASURES['YLLs']:
+        rr_data = rr[(rr.cause_id == cid) & (rr.morbidity == 0)]
```
collijk: add `& (rr.mortality == 1)`

src/vivarium_inputs/validation/raw.py
```diff
@@ -862,3 +867,25 @@ def check_cause_age_restrictions(entity: Cause) -> None:
...
+    else:
+        rr_data = rr[(rr.cause_id == cid) & (rr.morbidity == 1)]
+    missing_age_ids = set(rr_data.age_group_id)-set(paf.age_group_id)
+    if missing_age_ids and not np.allclose(rr_data[missing_age_ids][DRAW_COLUMNS], 0):
```
collijk: Shouldn't the check be for rr data != 1?

collijk: I'm also not sure this is a good check.  We could have relative risks in an age group but no exposure, leading to 0 paf.

collijk: If there is exposure and relative risk, there should be a paf.  If there is no paf, either rr is 1 or there is no exposure.

src/vivarium_inputs/validation/raw.py
```diff
@@ -862,3 +867,25 @@ def check_cause_age_restrictions(entity: Cause) -> None:
...
 
+
+def check_paf_rr_age_groups(paf: pd.DataFrame, rr: pd.DataFrame) -> None:
+    """Raise an error if relative risk includes any age_group_ids with non
+    trivial data. But paf still can have an extra data while there is no
+    relative risk matching with.
+    """
+    measure_map = {MEASURES['YLLs']: 'YLLs', MEASURES['YLDs']: 'YLDs'}
+    cid = paf.cause_id.unique()[0]
+    measure = paf.measure_id.unique()[0]
```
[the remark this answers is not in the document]
michaella21: nope before call this function, paf is groupby both cause_id and measure_id, so whenever it hits this function there should be exactly one cause id and measure id

src/vivarium_inputs/validation/raw.py
michaella21: yes, it's the reason why we are not using check_age_restrictions since that will raise an error for both cases of missing age groups and extra age groups

src/vivarium_inputs/validation/raw.py
kfwilson: you guys have talked about this a bunch but I'm still a little confused/think maybe I missed something. It's never going to be the case that we have a YLL paf and an RR that just has mort = morb = 1 for all rows?

src/vivarium_inputs/validation/raw.py
michaella21: We do check whether age groups of relative risk have exposure data, i.e., it should be guaranteed that we have both exposure and rr data for the age groups ids of RR already. Do we still need a separate check for the second comment?

michaella21: In other words, it should be guaranteed that we have both exposure and RR for a set of age groups and here we are checking whether we have all the pafs for those same age groups. If there is no paf for those groups that RR and exposure exist, we raise an error here. so please specify what kind of checks should be done more in details. @collijk

src/vivarium_inputs/extract.py
```diff
@@ -129,7 +129,7 @@ def extract_relative_risk(entity, location_id: int) -> pd.DataFrame:
     if entity.kind == 'risk_factor':
         data = gbd.get_relative_risk(entity.gbd_id, location_id)
         data = filter_to_most_detailed_causes(data)
-    else:  # coverage_gap
```
collijk: Think you accidentally dropped a space

src/vivarium_inputs/validation/raw.py
```diff
@@ -123,6 +123,7 @@ def validate_raw_data(data: pd.DataFrame, entity: ModelableEntity,
         'exposure_distribution_weights': validate_exposure_distribution_weights,
         'relative_risk': validate_relative_risk,
         'population_attributable_fraction': validate_population_attributable_fraction,
+        'etiology_population_attributable_fraction': validate_population_attributable_fraction,
```
collijk: While this is clever, I think it's going to be more trouble than it's worth.  There are a number of shared validations between risk paf and etiology paf, but also a bunch of branches in the validator we could get rid of if we split these into separate functions.

src/vivarium_inputs/validation/raw.py
```diff
@@ -867,3 +876,25 @@ def check_cause_age_restrictions(entity: Cause) -> None:
...
+    """
+    measure_map = {MEASURES['YLLs']: 'YLLs', MEASURES['YLDs']: 'YLDs'}
+    cid = paf.cause_id.unique()[0]
+    measure = paf.measure_id.unique()[0]
+
+    if measure == MEASURES['YLLs']:
+        rr_data = rr[(rr.cause_id == cid) & (rr.morbidity == 0) & (rr.mortality == 1)]
+    else:
+        rr_data = rr[(rr.cause_id == cid) & (rr.morbidity == 1)]
+    missing_age_ids = set(rr_data.age_group_id)-set(paf.age_group_id)
```
collijk: So a few things that concern me about this test:

It depends on the rr and paf data representing missingness in the same way:

I remember that most of the exposure/rr data we pull uses true missing rows (rather than filled with zeros/ones) to represent no possible exposure in the data.  Do we know if that's universally true though?  

How does the paf data represent missingness?  I would think, since it's a final gbd output, that it comes with full age ranges with zero values set in appropriate places.

If either the rr data is inconsistent or the paf data represents missingness with zeros rather than deleted rows, doing a check on the age groups is not going to work correctly.  

The by age value checks we want are:
1. if there is nonzero exposure to the risk in an age bin and there are non-trivial rr values for the correct measure associated with that exposure, then there must be a nonzero paf, so raise an error
2. If there are nonzero paf values for age bins with zero exposure (or non-existent rows) or there are nonzero paf values for age bins with exposure but no rr values for the correct measure, then warn.  This is a situation where some age bins of paf were calculated directly, extrapolated, or something else we don't understand.

collijk: On the other side of this is cause restrictions.  If there are nonzero paf values for YLDs for a cause outside the cause yld restrictions, then we should raise an error.  Same for YLLs.

michaella21:
on hold until I finish the James' last comments :(

michaella21:
close it since there was a way too many changes in feature branch, will open again once it's ready

PR #133, opened 2019-02-04
Feature/refactor estimation years

collijk:
This handles half of the estimation year refactor (the raw validator part).  Sim validator section coming next.

Also addressed leftover constants from the last pr.

### DIFF
--- a/src/vivarium_inputs/validation/utilities.py
+++ b/src/vivarium_inputs/validation/utilities.py
@@ -16,11 +16,11 @@ def check_years(data: pd.DataFrame, year_type: str, estimation_years: pd.Series)
 
     Parameters
     ----------
-    data :
+    data
         Dataframe containing 'year_id' column.
-    year_type :
-        String 'annual' or 'binned' indicating expected year range.
-    estimation_years :
+    year_type
+        'annual', 'binned', or 'either' indicating expected year range.
+    estimation_years
         The set of years the data should be estimated for.
 
     Raises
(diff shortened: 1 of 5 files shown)


src/vivarium_inputs/validation/utilities.py
```diff
@@ -11,23 +11,17 @@
...
 
 
-def check_years(data: pd.DataFrame, year_type: str, error: bool = True):
+def check_years(data: pd.DataFrame, year_type: str, estimation_years: pd.Series) -> None:
     """Check that years in passed data match expected range based on type.
 
     Parameters
     ----------
-    data
+    data :
```
kfwilson: I don't think you do the : if you're not specifying type - at least, that's what the numpy docstring guide I've been following says (https://numpydoc.readthedocs.io/en/latest/format.html). Let me know if we want to always specify the colon though and I can update the doc strings I'm writing

src/vivarium_inputs/validation/utilities.py
```diff
@@ -11,23 +11,17 @@
...
         Dataframe containing 'year_id' column.
-    year_type
+    year_type :
         String 'annual' or 'binned' indicating expected year range.
```
kfwilson: This should also include 'either' now right?

src/vivarium_inputs/validation/utilities.py
collijk: Ah, no, we should do the numpy thing.  I've been putting colons in all our docs forever though :-1:

src/vivarium_inputs/validation/utilities.py
collijk: yep

PR #134, opened 2019-02-05
Add validation context, sim validator estimation year refactor

collijk:
Parts of this PR are provisional.  In particular, core's dependence on `gbd`.  I just needed a shim to split the work into separate PRs.  Things to pay attention to:

- Every function in core needs to have an entry in the top level dictionary.  I missed it when exposure age groups came in.
- ``get_demographic_dimensions`` take an additional kwarg.  There are two use cases for core functions: they get called from the top level ``get_data`` or they get called directly by other core functions.  I think the kwarg thing is fine beyond the standard entity and location id arguments given defaults that would produce data that makes sense to an external user.
- The validation context.  This is a light wrapper around a dict.  I'll probably switch it in the next pr to actually inherit from ``UserDict`` if that seems reasonably easy.  It's constructed from context args passed to the main validate for simulation function.  It might be nice to have this as more of a global switch or as a context manager directly.  I'm thinking for the use case where someone wants to build an artifact with different age or year bins but still validate things well.  Could use some thoughts on this.

### DIFF
--- a/src/vivarium_inputs/core.py
+++ b/src/vivarium_inputs/core.py
@@ -6,11 +6,9 @@ from gbd_mapping import Cause, Sequela, RiskFactor, CoverageGap, Etiology, Covar
 import pandas as pd
 import numpy as np
 
-from vivarium_inputs import utilities, extract
+from vivarium_inputs import utilities, extract, utility_data
 from vivarium_inputs.globals import InvalidQueryError, DEMOGRAPHIC_COLUMNS, MEASURES, SEXES, Population, gbd
 from vivarium_inputs.mapping_extension import AlternativeRiskFactor, HealthcareEntity, HealthTechnology
-
-
 
 
 def get_data(entity, measure: str, location: str):
@@ -41,7 +39,6 @@ def get_data(entity, measure: str, location: str):
         'structure': (get_structure, ('population',)),
         'theoretical_minimum_risk_life_expectancy': (get_theoretical_minimum_risk_life_expectancy, ('population',)),
         'age_bins': (get_age_bins, ('population',)),
-        'estimation_years': (get_estimation_years, ('population',)),
         'demographic_dimensions': (get_demographic_dimensions, ('population',))
 
     }
@@ -338,16 +335,9 @@ def get_age_bins(entity: Population, location_id: int) -> pd.DataFrame:
     return age_bins
 
 
-def get_estimation_years(entity: Population, location_id: int) -> pd.DataFrame:
-    estimation_years = extract.extract_data(entity, 'estimation_years', location_id)
-    estimation_years = pd.DataFrame({'year_start': range(min(estimation_years), max(estimation_years) + 1)})
-    estimation_years['year_end'] = estimation_years['year_start'] + 1
-    return estimation_years
-
-
 def get_demographic_dimensions(entity: Population, location_id: int, draws: bool = False) -> pd.DataFrame:
     ages = gbd.get_age_group_id()
-    estimation_years = extract.extract_data(entity, 'estimation_years', location_id)
+    estimation_years = utility_data.get_estimation_years()
     years = range(min(estimation_years), max(estimation_years) + 1)
     sexes = [SEXES['Male'], SEXES['Female']]
     location = [location_id]
--- a/src/vivarium_inputs/core.py
+++ b/src/vivarium_inputs/core.py
@@ -25,6 +25,7 @@ def get_data(entity, measure: str, location: str):
         'case_fatality': (get_case_fatality, ('cause',)),
         # Risk-like measures
         'exposure': (get_exposure, ('risk_factor', 'coverage_gap', 'alternative_risk_factor',)),
+        'exposure_age_groups': (get_exposure_age_groups, ('risk_factor', 'alternative_risk_factor',)),
         'exposure_standard_deviation': (get_exposure_standard_deviation, ('risk_factor', 'alternative_risk_factor')),
         'exposure_distribution_weights': (get_exposure_distribution_weights, ('risk_factor', 'alternative_risk_factor')),
         'relative_risk': (get_relative_risk, ('risk_factor', 'coverage_gap')),
@@ -39,7 +40,9 @@ def get_data(entity, measure: str, location: str):
         'structure': (get_structure, ('population',)),
         'theoretical_minimum_risk_life_expectancy': (get_theoretical_minimum_risk_life_expectancy, ('population',)),
         'age_bins': (get_age_bins, ('population',)),
+        'estimation_years': (get_estimation_years, ('population',)),
         'demographic_dimensions': (get_demographic_dimensions, ('population',))
+
     }
 
     if measure not in measure_handlers:
@@ -171,6 +174,11 @@ def get_exposure(entity: Union[RiskFactor, AlternativeRiskFactor, CoverageGap], 
         data = utilities.normalize(data, fill_value=0)
     data = utilities.reshape(data, to_keep=DEMOGRAPHIC_COLUMNS + ['parameter'])
     return data
+
+
+def get_exposure_age_groups(entity: RiskFactor, location_id: int):
+    exposure = extract.extract_data(entity, 'exposure', location_id)
+    return set(exposure.age_group_id)
 
 
 def get_exposure_standard_deviation(entity: Union[RiskFactor, AlternativeRiskFactor], location_id: int) -> pd.DataFrame:
@@ -329,12 +337,15 @@ def get_age_bins(entity: Population, location_id: int) -> pd.DataFrame:
     return age_bins
 
 
+def get_estimation_years(entity: Population, location_id: int) -> pd.DataFrame:
+    estimation_years = extract.extract_data(entity, 'estimation_years', location_id)
+    estimation_years = pd.DataFrame({'year_start': range(min(estimation_years), max(estimation_years) + 1)})
+    estimation_years['year_end'] = estimation_years['year_start'] + 1
+    return estimation_years
+
+
 def get_demographic_dimensions(entity: Population, location_id: int) -> pd.DataFrame:
     demographic_dimensions = utilities.get_demographic_dimensions(location_id)
     demographic_dimensions = utilities.normalize(demographic_dimensions)
     return demographic_dimensions
 
-
-def get_exposure_age_groups(entity: RiskFactor, location_id: int):
-    exposure = extract.extract_data(entity, 'exposure', location_id)
-    return set(exposure.age_group_id)
--- a/src/vivarium_inputs/core.py
+++ b/src/vivarium_inputs/core.py
@@ -7,9 +7,10 @@ import pandas as pd
 import numpy as np
 
 from vivarium_inputs import utilities, extract
+from vivarium_inputs.globals import InvalidQueryError, DEMOGRAPHIC_COLUMNS, MEASURES, SEXES, Population, gbd
 from vivarium_inputs.mapping_extension import AlternativeRiskFactor, HealthcareEntity, HealthTechnology
 
-from .globals import InvalidQueryError, DEMOGRAPHIC_COLUMNS, MEASURES, Population
+
 
 
 def get_data(entity, measure: str, location: str):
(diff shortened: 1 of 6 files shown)


src/vivarium_inputs/core.py
```diff
@@ -334,12 +341,27 @@ def get_age_bins(entity: Population, location_id: int) -> pd.DataFrame:
     return age_bins
 
 
-def get_demographic_dimensions(entity: Population, location_id: int) -> pd.DataFrame:
-    demographic_dimensions = utilities.get_demographic_dimensions(location_id)
+def get_estimation_years(entity: Population, location_id: int) -> pd.DataFrame:
```
halfhorst: estimation year has bothered me for a bit -- sometimes it isn't clear if you are getting annual or binned, and sometimes it is labeled. Should we make a push  to label it so it's clear when reading code? Here it looks like it is going from binned to annual

src/vivarium_inputs/core.py
```diff
@@ -334,12 +341,27 @@ def get_age_bins(entity: Population, location_id: int) -> pd.DataFrame:
...
+    estimation_years = extract.extract_data(entity, 'estimation_years', location_id)
+    estimation_years = pd.DataFrame({'year_start': range(min(estimation_years), max(estimation_years) + 1)})
+    estimation_years['year_end'] = estimation_years['year_start'] + 1
+    return estimation_years
+
+
+def get_demographic_dimensions(entity: Population, location_id: int, draws: bool = False) -> pd.DataFrame:
+    ages = gbd.get_age_group_id()
+    estimation_years = extract.extract_data(entity, 'estimation_years', location_id)
+    years = range(min(estimation_years), max(estimation_years) + 1)
```
michaella21: if `get_estimation_years` is changed from binned to annual, we don't need to make range again from its min/max?

src/vivarium_inputs/core.py
collijk: Maybe in the docstring?  This is consistent with the rest of the data in our architecture.  It's raw binned data at the extract level and full rank data after it passes through core/normalize.

kfwilson:
> Parts of this PR are provisional. In particular, core's dependence on `gbd`. I just needed a shim to split the work into separate PRs. Things to pay attention to:
> 
> * Every function in core needs to have an entry in the top level dictionary.  I missed it when exposure age groups came in.
> * `get_demographic_dimensions` take an additional kwarg.  There are two use cases for core functions: they get called from the top level `get_data` or they get called directly by other core functions.  I think the kwarg thing is fine beyond the standard entity and location id arguments given defaults that would produce data that makes sense to an external user.
> * The validation context.  This is a light wrapper around a dict.  I'll probably switch it in the next pr to actually inherit from `UserDict` if that seems reasonably easy.  It's constructed from context args passed to the main validate for simulation function.  It might be nice to have this as more of a global switch or as a context manager directly.  I'm thinking for the use case where someone wants to build an artifact with different age or year bins but still validate things well.  Could use some thoughts on this.


I'm not sure I totally understand your second point about get_demographic_dimensions. Are you asking about whether it would make sense to take kwargs so that the user could provide the age/year/etc. bins to build demographic dimensions over? Is the use case for this the one you mention in your next point about someone wanting to build age/year bins different than standard gbd?

I really like the validation context. I think it makes things a lot cleaner and clearer. I also think the idea of having it not hard-wired into gbd so someone could use different age/year bins would be a really handy feature.

src/vivarium_inputs/core.py
kfwilson: Yeah I definitely think clarifying in the docstring would be helpful. I get confused because we have an identically named function in gbd_access that gives the binned data

src/vivarium_inputs/core.py
```diff
@@ -334,12 +341,27 @@ def get_age_bins(entity: Population, location_id: int) -> pd.DataFrame:
     return age_bins
 
 
-def get_demographic_dimensions(entity: Population, location_id: int) -> pd.DataFrame:
-    demographic_dimensions = utilities.get_demographic_dimensions(location_id)
+def get_estimation_years(entity: Population, location_id: int) -> pd.DataFrame:
+    estimation_years = extract.extract_data(entity, 'estimation_years', location_id)
+    estimation_years = pd.DataFrame({'year_start': range(min(estimation_years), max(estimation_years) + 1)})
+    estimation_years['year_end'] = estimation_years['year_start'] + 1
```
kfwilson: should this happen here or should it be set to 'year_id' and then the scrub gbd conventions in interface would add the year_start, year_end?

halfhorst:
> > Parts of this PR are provisional. In particular, core's dependence on `gbd`. I just needed a shim to split the work into separate PRs. Things to pay attention to:
> > 
> > * Every function in core needs to have an entry in the top level dictionary.  I missed it when exposure age groups came in.
> > * `get_demographic_dimensions` take an additional kwarg.  There are two use cases for core functions: they get called from the top level `get_data` or they get called directly by other core functions.  I think the kwarg thing is fine beyond the standard entity and location id arguments given defaults that would produce data that makes sense to an external user.
> > * The validation context.  This is a light wrapper around a dict.  I'll probably switch it in the next pr to actually inherit from `UserDict` if that seems reasonably easy.  It's constructed from context args passed to the main validate for simulation function.  It might be nice to have this as more of a global switch or as a context manager directly.  I'm thinking for the use case where someone wants to build an artifact with different age or year bins but still validate things well.  Could use some thoughts on this.
> 
> I'm not sure I totally understand your second point about get_demographic_dimensions. Are you asking about whether it would make sense to take kwargs so that the user could provide the age/year/etc. bins to build demographic dimensions over? Is the use case for this the one you mention in your next point about someone wanting to build age/year bins different than standard gbd?
> 
> I really like the validation context. I think it makes things a lot cleaner and clearer. I also think the idea of having it not hard-wired into gbd so someone could use different age/year bins would be a really handy feature.

I think he's just pointing out that before get_demog, the function signatures were completely consistent, taking entity and location. get_demographic_dimensions diverges from this but that's fine as long as the kwarg default is sensible

src/vivarium_inputs/core.py
collijk: I've pulled out all references (save for what's in normalize, which I'll deal with in a different PR) to the ``gbd`` function.  The end point is no usage of ``gbd`` outside ``extract``.

src/vivarium_inputs/core.py
collijk: This is a good question.  Let me think about it.

kfwilson:
> > > Parts of this PR are provisional. In particular, core's dependence on `gbd`. I just needed a shim to split the work into separate PRs. Things to pay attention to:
> > > 
> > > * Every function in core needs to have an entry in the top level dictionary.  I missed it when exposure age groups came in.
> > > * `get_demographic_dimensions` take an additional kwarg.  There are two use cases for core functions: they get called from the top level `get_data` or they get called directly by other core functions.  I think the kwarg thing is fine beyond the standard entity and location id arguments given defaults that would produce data that makes sense to an external user.
> > > * The validation context.  This is a light wrapper around a dict.  I'll probably switch it in the next pr to actually inherit from `UserDict` if that seems reasonably easy.  It's constructed from context args passed to the main validate for simulation function.  It might be nice to have this as more of a global switch or as a context manager directly.  I'm thinking for the use case where someone wants to build an artifact with different age or year bins but still validate things well.  Could use some thoughts on this.
> > 
> > 
> > I'm not sure I totally understand your second point about get_demographic_dimensions. Are you asking about whether it would make sense to take kwargs so that the user could provide the age/year/etc. bins to build demographic dimensions over? Is the use case for this the one you mention in your next point about someone wanting to build age/year bins different than standard gbd?
> > I really like the validation context. I think it makes things a lot cleaner and clearer. I also think the idea of having it not hard-wired into gbd so someone could use different age/year bins would be a really handy feature.
> 
> I think he's just pointing out that before get_demog, the function signatures were completely consistent, taking entity and location. get_demographic_dimensions diverges from this but that's fine as long as the kwarg default is sensible

Oh ok that makes sense. I'm not 100% convinced by this tbh because I feel like it makes the naming of the function ambiguous - if you pass draws=True, you're not actually getting just demographic dimensions anymore. But since we're only going to expose the top level get_data function, it doesn't really matter.

collijk:
> > > > Parts of this PR are provisional. In particular, core's dependence on `gbd`. I just needed a shim to split the work into separate PRs. Things to pay attention to:
> > > > 
> > > > * Every function in core needs to have an entry in the top level dictionary.  I missed it when exposure age groups came in.
> > > > * `get_demographic_dimensions` take an additional kwarg.  There are two use cases for core functions: they get called from the top level `get_data` or they get called directly by other core functions.  I think the kwarg thing is fine beyond the standard entity and location id arguments given defaults that would produce data that makes sense to an external user.
> > > > * The validation context.  This is a light wrapper around a dict.  I'll probably switch it in the next pr to actually inherit from `UserDict` if that seems reasonably easy.  It's constructed from context args passed to the main validate for simulation function.  It might be nice to have this as more of a global switch or as a context manager directly.  I'm thinking for the use case where someone wants to build an artifact with different age or year bins but still validate things well.  Could use some thoughts on this.
> > > 
> > > 
> > > I'm not sure I totally understand your second point about get_demographic_dimensions. Are you asking about whether it would make sense to take kwargs so that the user could provide the age/year/etc. bins to build demographic dimensions over? Is the use case for this the one you mention in your next point about someone wanting to build age/year bins different than standard gbd?
> > > I really like the validation context. I think it makes things a lot cleaner and clearer. I also think the idea of having it not hard-wired into gbd so someone could use different age/year bins would be a really handy feature.
> > 
> > 
> > I think he's just pointing out that before get_demog, the function signatures were completely consistent, taking entity and location. get_demographic_dimensions diverges from this but that's fine as long as the kwarg default is sensible
> 
> Oh ok that makes sense. I'm not 100% convinced by this tbh because I feel like it makes the naming of the function ambiguous - if you pass draws=True, you're not actually getting just demographic dimensions anymore. But since we're only going to expose the top level get_data function, it doesn't really matter.

Yeah, this is actually an implementation issue rather than anything else.  If I wanted to grab demography, then add a draw column, I'd have to make 1000 copies of the demographic block first.  Which is super gross.

kfwilson:
> > > > > Parts of this PR are provisional. In particular, core's dependence on `gbd`. I just needed a shim to split the work into separate PRs. Things to pay attention to:
> > > > > 
> > > > > * Every function in core needs to have an entry in the top level dictionary.  I missed it when exposure age groups came in.
> > > > > * `get_demographic_dimensions` take an additional kwarg.  There are two use cases for core functions: they get called from the top level `get_data` or they get called directly by other core functions.  I think the kwarg thing is fine beyond the standard entity and location id arguments given defaults that would produce data that makes sense to an external user.
> > > > > * The validation context.  This is a light wrapper around a dict.  I'll probably switch it in the next pr to actually inherit from `UserDict` if that seems reasonably easy.  It's constructed from context args passed to the main validate for simulation function.  It might be nice to have this as more of a global switch or as a context manager directly.  I'm thinking for the use case where someone wants to build an artifact with different age or year bins but still validate things well.  Could use some thoughts on this.
> > > > 
> > > > 
> > > > I'm not sure I totally understand your second point about get_demographic_dimensions. Are you asking about whether it would make sense to take kwargs so that the user could provide the age/year/etc. bins to build demographic dimensions over? Is the use case for this the one you mention in your next point about someone wanting to build age/year bins different than standard gbd?
> > > > I really like the validation context. I think it makes things a lot cleaner and clearer. I also think the idea of having it not hard-wired into gbd so someone could use different age/year bins would be a really handy feature.
> > > 
> > > 
> > > I think he's just pointing out that before get_demog, the function signatures were completely consistent, taking entity and location. get_demographic_dimensions diverges from this but that's fine as long as the kwarg default is sensible
> > 
> > 
> > Oh ok that makes sense. I'm not 100% convinced by this tbh because I feel like it makes the naming of the function ambiguous - if you pass draws=True, you're not actually getting just demographic dimensions anymore. But since we're only going to expose the top level get_data function, it doesn't really matter.
> 
> Yeah, this is actually an implementation issue rather than anything else. If I wanted to grab demography, then add a draw column, I'd have to make 1000 copies of the demographic block first. Which is super gross.

Oh ok. I didn't realize that if you give it draws, it gives you back a data frame long on draws. That actually makes me even more uncomfortable with the name because then if you pass draws=True, not only are you not getting back just demographic dimensions, you're actually getting back 1000 copies of demographic dimensions. I understand why we need it but let's just make sure to be very clear in the doc string when we document core.

PR #139, opened 2019-02-05
Rearrange dependencies in the utilities.

collijk:
- This removes ``validation/utilities.py`` and replaces it with ``validation/shared.py``.  
- Utilities specific to a single validation stage have been moved into the specific validation model.
- Some utilities used in both validation and core have been moved to the main utilities module.


tests/validation/test_raw.py
```diff
@@ -27,3 +28,324 @@ def test_check_mort_morb_flags_fail(mort, morb, yld_only, yll_only, match):
...
+test_data = [(pd.DataFrame({'a': [0, 1], 'b': [2, 20]}), 0, 'lower', ['a', 'b'], True),
+             (pd.DataFrame({'a': [0, 1], 'b': [2, 20]}), -1, 'lower', ['a', 'b'], False),
+             (pd.DataFrame({'a': [0], 'b': [10], 'c': [100]}), 10, 'lower', ['b', 'c'], True),
+             (pd.DataFrame({'a': [0, 1], 'b': [2, 20]}), 20, 'upper', ['a', 'b'], True),
+             (pd.DataFrame({'a': [0, 1], 'b': [2, 20]}, index=[0, 1]), pd.Series([0, 1], index=[0, 1]), 'upper',
+              ['a'], True)]
+
+
+@pytest.mark.parametrize('data, boundary, boundary_type, val_cols, inclusive', test_data)
+def test_check_value_columns_boundary_pass(data, boundary, boundary_type, val_cols, inclusive):
```
kfwilson: this seems like it should be in a test_shared file?

collijk: yep

PR #141, opened 2019-02-05
need measure id, not string

halfhorst:
This bug appears to have come back

### DIFF
--- a/src/vivarium_inputs/validation/raw.py
+++ b/src/vivarium_inputs/validation/raw.py
@@ -7,7 +7,7 @@ import numpy as np
 from gbd_mapping import (ModelableEntity, Cause, Sequela, RiskFactor,
                          Etiology, Covariate, CoverageGap, causes)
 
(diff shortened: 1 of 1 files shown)


src/vivarium_inputs/validation/raw.py
```diff
@@ -965,14 +965,13 @@ def validate_population_attributable_fraction(data: pd.DataFrame, entity: Union[
 
     for c_id in data.cause_id:
         cause = [c for c in causes if c.gbd_id == c_id][0]
-        if cause.restrictions.yld_only and (data.measure_id == 'YLLs').any():
+        if cause.restrictions.yld_only and (data.measure_id == MEASURES['YLLs']).any():
```
michaella21: it actually should be data[(data.measure_id==..) & (data.cause_id == c_id)] also in #971, you can fix it or just leave it since it's already in mine :)

collijk: I'd say go ahead and fix it so we can get the artifact for the BMI model

halfhorst: I've got this on my csu hack branch. I'll close this and sit tight until michelle's PR comes in

PR #143, opened 2019-02-05
util validators

kfwilson:
here's the last set of doc strings (pending paf) for the raw validators - this includes all the util functions that were missing doc strings

+ cleaning up some previous doc strings because grammar

### DIFF
--- a/src/vivarium_inputs/validation/raw.py
+++ b/src/vivarium_inputs/validation/raw.py
@@ -1012,9 +1012,12 @@ def validate_estimate(data: pd.DataFrame, entity: Covariate,
     check_years(data, 'annual', estimation_years)
     check_location(data, location_id)
 
-    if entity.by_age and not set(data.age_group_id).intersection(set(gbd.get_age_group_id())):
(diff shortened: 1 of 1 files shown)


src/vivarium_inputs/validation/raw.py
```diff
@@ -1309,9 +1372,26 @@ def check_covariate_sex_restriction(data: pd.DataFrame, by_sex: bool):
 
 
 def check_covariate_age_restriction(data: pd.DataFrame, by_age: bool) -> None:
-    """ Because covariate age restrictions are simply by_age or not rather than
```
collijk: Do you think it's worth having separate functions for these rather than just having the validation code inside covariate?

kfwilson: Probably not. It's probably easier to understand to just have them in the covariate validator rather than having to jump here to see what specially is being check for covariate restrictions. I'll move them inside the covariate validator

kfwilson: combined them with the main covariate validation. It actually cleaned things up and removed some redundant checks in estimate validator

PR #144, opened 2019-02-05
Lift out age bins

collijk:


halfhorst:
I can't undo my approval while I wait for updates :laughing:

src/vivarium_inputs/validation/raw.py
```diff
@@ -377,6 +377,8 @@ def validate_incidence(data: pd.DataFrame, entity: Union[Cause, Sequela],
         Location to which the data should pertain.
     estimation_years
         Expected set of years, used to check the `year_id` column in `data`.
+    age_group_ids
+        Expected set of age group ids.
```
kfwilson: I think this should be list of possible age ids as in the doc strings for below methods

src/vivarium_inputs/validation/raw.py
```diff
@@ -707,7 +719,7 @@ def validate_exposure(data: pd.DataFrame, entity: Union[RiskFactor, CoverageGap,
         male_expected = not restrictions.female_only
         female_expected = not restrictions.male_only
 
-        cats.apply(check_age_group_ids, None, None)
+        cats.apply(check_age_group_ids, age_group_ids, None, None)
```
kfwilson: I can't comment on it because it hasn't changed, but I think you need this additional age_group_ids argument on line 736 below

src/vivarium_inputs/validation/raw.py
```diff
@@ -923,7 +942,7 @@ def validate_relative_risk(data: pd.DataFrame, entity: Union[RiskFactor, Coverag
                 start, end = cause.restrictions.yld_age_group_id_start, cause.restrictions.yld_age_group_id_end
             else:  # morb = 0 , mort = 1
                 start, end = cause.restrictions.yll_age_group_id_start, cause.restrictions.yll_age_group_id_end
-            check_age_restrictions(g, start, end, error=False)
+            check_age_restrictions(g, age_group_ids, start, end, error=False)
```
kfwilson: I can't comment because not changed again but I think you need age_group_ids to be passed to the apply call to check_age_group_ids call on line 931 above

src/vivarium_inputs/validation/raw.py
```diff
@@ -1555,8 +1581,8 @@ def check_age_group_ids(data: pd.DataFrame, restriction_start: float = None, res
         or they don't make up a contiguous block.
 
     """
-    all_ages = set(gbd.get_age_group_id())
-    restriction_ages = set(get_restriction_age_ids(restriction_start, restriction_end))
+    all_ages = set()
```
kfwilson: should this be `all_ages = set(age_group_ids)`?

src/vivarium_inputs/core.py
```diff
@@ -332,12 +338,12 @@ def get_theoretical_minimum_risk_life_expectancy(entity: Population, location_id
...
 
 def get_age_bins(entity: Population, location_id: int) -> pd.DataFrame:
-    age_bins = utilities.get_age_bins()[['age_group_name', 'age_group_start', 'age_group_end']]
+    age_bins = utility_data.get_age_bins()[['age_group_name', 'age_group_start', 'age_group_end']]
     return age_bins
 
 
 def get_demographic_dimensions(entity: Population, location_id: int, draws: bool = False) -> pd.DataFrame:
-    ages = gbd.get_age_group_id()
+    ages = utility_data.get_age_bins().age_group_id
```
kfwilson: why are you using utility_data.get_age_bins() instead of utility_data.get_age_group_ids()?

src/vivarium_inputs/extract.py
```diff
@@ -6,35 +6,35 @@
...
+        'remission': (extract_remission, (get_estimation_years, get_age_group_ids)),
+        'deaths': (extract_deaths, (get_estimation_years, get_age_group_ids, extract_structure)),
         # Risk-like measures
-        'exposure': (extract_exposure, (get_estimation_years, )),
+        'exposure': (extract_exposure, (get_estimation_years, get_age_group_ids)),
         'exposure_standard_deviation': (extract_exposure_standard_deviation,
                                         (extract_exposure, get_estimation_years)),
         'exposure_distribution_weights': (extract_exposure_distribution_weights, ()),
-        'relative_risk': (extract_relative_risk, (extract_exposure, get_estimation_years)),
+        'relative_risk': (extract_relative_risk, (extract_exposure, get_estimation_years, get_age_group_ids)),
```
kfwilson: this is minor, but I think it might be helpful to standardize the order we put the extra data in. e.g., we always pass estimation years then ages then whatever else. just because I could see myself getting confused and not realizing that the raw validators are expecting data in a certain order. and because it would make the functions in raw validators a little more uniform

PR #146, opened 2019-02-06
Feature/handle protective

halfhorst:
This PR handles the protective relationship between risk factors and causes. It introduces a global dict mapping risk_factor names to cause entities (and also lifts PAFs constants to globals at the top in raw.py). This is used when validating PAF values to alter the acceptable minimum, in both the raw and the simulation validators.

Incidentally, breast cancer and BMI have a contentious relationship, the protective effect is potentially real. I found that interesting.

It would be nice to make the gbd entities hashable so I could stick the risk_factor in directly as the key, but that's for another time.


collijk:
Michelle to resolve in her paf update

PR #153, opened 2019-02-07
Shift towards context usage in the raw validators

collijk:
I should have done this to start.

### DIFF
--- a/src/vivarium_inputs/validation/raw.py
+++ b/src/vivarium_inputs/validation/raw.py
@@ -659,7 +659,7 @@ def validate_deaths(data: pd.DataFrame, entity: Cause, context: RawValidationCon
     population = context['population']
     population = population[(population.age_group_id.isin(data.age_group_id.unique()))
                             & (population.year_id.isin(data.year_id.unique()))
-                            & (population.sex_id != SEXES['Combined'])].set_index(idx_cols).population
+                            & (population.sex_id != context['sexes']['Combined'])].set_index(idx_cols).population
     check_value_columns_boundary(data.set_index(idx_cols), population, 'upper',
                                  value_columns=DRAW_COLUMNS, inclusive=True, error=None)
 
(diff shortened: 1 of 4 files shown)


src/vivarium_inputs/validation/raw.py
```diff
@@ -1316,6 +1279,33 @@ def check_paf_types(entity: Union[Etiology, RiskFactor]) -> None:
                       f'{entity.kind} {entity.name} may be outside expected range [0, 1].')
 
 
+def check_cause_age_restrictions_sets(entity: Cause) -> None:
```
kfwilson: why did you move this out of the utilities section?

src/vivarium_inputs/validation/raw.py
```diff
@@ -644,23 +640,23 @@ def validate_deaths(data: pd.DataFrame, entity: Cause, location_id: int, populat
...
+    check_sex_ids(data, context, male_expected, female_expected)
 
-    check_age_restrictions(data, age_group_ids, restrictions.yll_age_group_id_start, restrictions.yll_age_group_id_end)
-    check_sex_restrictions(data, restrictions.male_only, restrictions.female_only)
+    check_age_restrictions(data, context, restrictions.yll_age_group_id_start, restrictions.yll_age_group_id_end)
+    check_sex_restrictions(data, context, restrictions.male_only, restrictions.female_only)
 
     check_value_columns_boundary(data, 0, 'lower', value_columns=DRAW_COLUMNS, inclusive=True, error=DataAbnormalError)
     idx_cols = ['age_group_id', 'year_id', 'sex_id']
+    population = context['population']
```
kfwilson: for filtering pop on line 662 below, you use the sexes in the validation context everywhere else but here is the globals sexes

src/vivarium_inputs/validation/raw.py
collijk: Moved from validation utilities to check metadata utilities.

PR #158, opened 2019-02-07
support relative_risk_scalar in artifact loaders

halfhorst:
This is the second step to getting relative_risk_scalar into an artifact -- identifying it as a key that can be loaded. In the map it is a scalar() type

### DIFF
--- a/src/vivarium_inputs/data_artifact/loaders.py
+++ b/src/vivarium_inputs/data_artifact/loaders.py
@@ -26,7 +26,8 @@ def loader(entity_key: EntityKey, location: str, modeled_causes: Set[str], all_m
             "getter": get_risk_data,
             "measures": ["affected_causes", "affected_risk_factors", "restrictions", "distribution",
                          "exposure_parameters", "categories", "tmred", "exposure", "exposure_standard_deviation",
-                         "relative_risk", "population_attributable_fraction", "exposure_distribution_weights"],
+                         "relative_risk", "relative_risk_scalar", "population_attributable_fraction",
+                         "exposure_distribution_weights"],
         },
         "alternative_risk_factor": {
             "mapping": alternative_risk_factors,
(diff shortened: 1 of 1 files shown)


src/vivarium_inputs/data_artifact/loaders.py
```diff
@@ -107,8 +108,8 @@ def get_cause_data(cause, measure, location, _):
 
 
 def get_risk_data(risk, measure, location, modeled_causes):
-    if measure in ["affected_causes", "affected_risk_factors", "restrictions",
-                   "distribution", "exposure_parameters", "categories", "tmred"]:
+    if measure in ["affected_causes", "affected_risk_factors", "restrictions", "distribution",
+                   "exposure_parameters", "categories", "tmred", "relative_risk_scalar"]:
```
collijk: There is no ``exposure_parameters`` anymore, right?

halfhorst: yeah totally right

halfhorst: not sure about alternative_risk_factors though

collijk: There's only three of them in the ``mapping_extension`` subpackage

halfhorst: i pulled categories and tmred from alternative_risk_factors because they weren't present as well

src/vivarium_inputs/data_artifact/loaders.py
```diff
@@ -25,15 +25,16 @@ def loader(entity_key: EntityKey, location: str, modeled_causes: Set[str], all_m
...
+                         "relative_risk", "relative_risk_scalar", "population_attributable_fraction",
+                         "exposure_distribution_weights"],
         },
         "alternative_risk_factor": {
             "mapping": alternative_risk_factors,
             "getter": get_risk_data,
-            "measures": ["affected_causes", "affected_risk_factors", "restrictions", "distribution",
-                         "exposure_parameters", "categories", "tmred", "exposure", "exposure_standard_deviation",
-                         "relative_risk", "population_attributable_fraction", "exposure_distribution_weights"],
+            "measures": ["affected_causes", "affected_risk_factors", "restrictions", "distribution", "exposure",
```
kfwilson: according to core, it doesn't look like we have paf or rr for alt risk factor?

halfhorst: oh I didn't look there. I'll just dump that, too!

PR #160, opened 2019-02-07
Docs/gbd 2017 tutorial

kfwilson:
here's a tutorial for using interface. It includes a table of available entity-measure pairs, but it's hard-coded (like the docstring for `get_measure` in interface) which I don't like because I already found that I'd forgotten a measure in the docstring so any clever ideas for auto-generating these?

### DIFF
--- a/docs/source/tutorials/pulling_data.rst
+++ b/docs/source/tutorials/pulling_data.rst
@@ -210,19 +210,16 @@ argument to fill the location dimension.
 
     funcs = {get_measure: {'entity': ModelableEntity, 'measure': str, 'location': str},
              get_population_structure: {'location': str},
-             get_theoretical_minimum_risk_life_expectancy: None,
-             get_age_bins: None,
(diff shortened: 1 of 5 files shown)


docs/source/tutorials/pulling_data.rst
```diff
@@ -0,0 +1,225 @@
...
+      draw location     sex  age_group_start  age_group_end  year_start  year_end     value
+    0    0    Kenya  Female              0.0       0.019178        1990      1991  0.032557
+    1    0    Kenya  Female              0.0       0.019178        1991      1992  0.031751
+    2    0    Kenya  Female              0.0       0.019178        1992      1993  0.031039
+    3    0    Kenya  Female              0.0       0.019178        1993      1994  0.030458
+    4    0    Kenya  Female              0.0       0.019178        1994      1995  0.030039
+
+The following table lists the measures available for each entity kind:
+
+.. list-table:: Available Entity-Measure Pairs
```
halfhorst: :pray:

PR #161, opened 2019-02-07
fix cov value check

kfwilson:
this is a little one - I realized when I went into sim validators to do docs that the cov value check wasn't quite checking what I wanted it to

### DIFF
--- a/src/vivarium_inputs/validation/sim.py
+++ b/src/vivarium_inputs/validation/sim.py
@@ -547,7 +547,7 @@ def check_covariate_values(data: pd.DataFrame):
 
     # allow the case where lower = mean = upper = 0 b/c of things like age
     # specific fertility rate where all estimates are 0 for young age groups
-    if np.all(data.value != 0) and not np.all(lower < mean < upper):
+    if not np.all(lower == mean == upper == 0 or lower < mean < upper):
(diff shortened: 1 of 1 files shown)


src/vivarium_inputs/validation/sim.py
```diff
@@ -547,7 +547,7 @@ def check_covariate_values(data: pd.DataFrame):
 
     # allow the case where lower = mean = upper = 0 b/c of things like age
     # specific fertility rate where all estimates are 0 for young age groups
-    if np.all(data.value != 0) and not np.all(lower < mean < upper):
+    if not np.all(lower == mean == upper == 0 or lower < mean < upper):
```
halfhorst: so, repeated back in plain english:

either *all* lower, upper mean are zero or *all* lower < mean < upper

kfwilson: Yep. Because for things like fertility rate, it'll be 0 for kids so I can't just check lower < mean < upper but before I was only doing that check if there were no 0s at all, and that's not really what I want

kfwilson: wait no that's not right. In english it should be all demographic groups either have upper, mean, lower are 0 or lower < mean < upper but not that all have to be 0 or all have to have lower < mean < upper

PR #169, opened 2019-02-09
Feature/update loaders

michaella21:
update the last bit of loaders

### DIFF
--- a/src/vivarium_inputs/data_artifact/loaders.py
+++ b/src/vivarium_inputs/data_artifact/loaders.py
@@ -34,7 +34,7 @@ def loader(entity_key: EntityKey, location: str, modeled_causes: Set[str], all_m
         "sequela": {
             "mapping": sequelae,
             "getter": get_sequela_data,
-            "measures": ["healthstate", "prevalence", "incidence", "disability_weight"],
+            "measures": ["healthstate", "prevalence", "incidence", "disability_weight", "birth_prevalence"],
         },
         "healthcare_entity": {
             "mapping": healthcare_entities,
@@ -165,7 +165,7 @@ def get_cause_metadata(entity, field):
     if field == "restrictions":
         data = entity.restrictions.to_dict()
     else:  # field in ["sequela", "etiologies"]:
-        if entity[field] is not None:
+        if entity[field] != ():
             data = [sub_entity.name for sub_entity in entity[field]]
         else:
             data = None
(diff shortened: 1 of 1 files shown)


src/vivarium_inputs/data_artifact/loaders.py
```diff
@@ -6,20 +6,16 @@
...
-from vivarium_inputs.mapping_extension import alternative_risk_factors
-
-
-CAUSE_BY_ID = {c.gbd_id: c for c in causes if c is not None}
-RISK_BY_ID = {r.gbd_id: r for r in risk_factors}
+from vivarium_inputs.mapping_extension import alternative_risk_factors, healthcare_entities, health_technologies
 
 
 def loader(entity_key: EntityKey, location: str, modeled_causes: Set[str], all_measures: bool = False):
     entity_data = {
```
kfwilson: do we not want to support sequela data in an artifact?

src/vivarium_inputs/data_artifact/loaders.py
```diff
@@ -40,21 +36,21 @@ def loader(entity_key: EntityKey, location: str, modeled_causes: Set[str], all_m
             "getter": get_sequela_data,
             "measures": ["healthstate", "prevalence", "incidence", "disability_weight"],
```
kfwilson: birth prevalence here too

src/vivarium_inputs/data_artifact/loaders.py
```diff
@@ -165,7 +165,7 @@ def get_cause_metadata(entity, field):
     if field == "restrictions":
         data = entity.restrictions.to_dict()
     else:  # field in ["sequela", "etiologies"]:
-        if entity[field] is not None:
+        if entity[field] != ():
```
kfwilson: I think you could just do if entity[field]? otherwise looks good!

src/vivarium_inputs/data_artifact/loaders.py
```diff
@@ -223,7 +174,7 @@ def get_cause_metadata(entity, field):
 
 def get_risk_metadata(risk, measure, modeled_causes):
     if measure in ["restrictions", "categories", "tmred"]:
-        if risk[measure] is not None:
```
kfwilson: I don't think I changed this though? I'm pretty sure tmred at least is still None

PR #170, opened 2019-02-09
Bugfix/rr warning

michaella21:
2 bugs 1 for rr warning and 1 for ensemble weights


src/vivarium_inputs/core.py
```diff
@@ -212,11 +212,11 @@ def get_exposure_distribution_weights(entity: Union[RiskFactor, AlternativeRiskF
             copied['age_group_id'] = age_id
             df.append(copied)
         data = pd.concat(df)
-
     distribution_cols = ['exp', 'gamma', 'invgamma', 'llogis', 'gumbel', 'invweibull', 'weibull',
                          'lnorm', 'norm', 'glnorm', 'betasr', 'mgamma', 'mgumbel']
-    id_cols = ['rei_id', 'location_id', 'sex_id', 'year_id', 'age_group_id', 'measure']
```
kfwilson: why aren't we normalizing by age?

kfwilson: also why are we only duplicating weights data for exposure groups for risk factors and not alt risk factors?

michaella21: ah no, i just got confused. I need to fix in risk distributions not here :(

kfwilson: ok but we still want to do this for alt risk factors too right?

src/vivarium_inputs/core.py
```diff
@@ -204,18 +204,17 @@ def get_exposure_standard_deviation(entity: Union[RiskFactor, AlternativeRiskFac
...
+    df = []
+    for age_id in set(extract.extract_data(entity, 'exposure', location_id).age_group_id):
+        copied = data.copy()
+        copied['age_group_id'] = age_id
+        df.append(copied)
+    data = pd.concat(df)
     distribution_cols = ['exp', 'gamma', 'invgamma', 'llogis', 'gumbel', 'invweibull', 'weibull',
                          'lnorm', 'norm', 'glnorm', 'betasr', 'mgamma', 'mgumbel']
-    id_cols = ['rei_id', 'location_id', 'sex_id', 'year_id', 'age_group_id', 'measure']
+    id_cols = ['location_id', 'sex_id', 'age_group_id', 'measure']
```
collijk: Does this work for both risks and alternative risks (that is are we consistent about the columns provided in both)?

michaella21: well, actually it worked for both, but i accidentally delete the rei_id, should have deleted just for year_id

collijk: :+1:

PR #172, opened 2019-02-09
filter exp sd data to restrictions

kfwilson:
we were filtering exp sd + exp dist weights data to the extracted version of exposure but exposure is later filtered to restriction age groups in core so there was an issue where if exp had extra age groups, we were passing those through in exp sd and then breaking in the age restrictions check in sim validators

this also now expands dist weights over years in addition to over ages

plus removes a reference to age group 235 in the check age restrictions warning message that's no longer relevant since michelle added that to the restrictions age groups


src/vivarium_inputs/validation/raw.py
```diff
@@ -1954,7 +1954,6 @@ def check_age_restrictions(data: pd.DataFrame, context: RawValidationContext,
         should_be_zero = data[data.age_group_id.isin(extra_age_groups)]
         if check_data_exist(should_be_zero, zeros_missing=True, value_columns=value_columns, error=False):
             warnings.warn(f'Data was only expected to contain values for age groups between ids '
```
collijk: Don't think this warning makes sense anymore.

kfwilson: whoops you're right. deleted too much. put back the age range part I didn't mean to delete

src/vivarium_inputs/core.py
```diff
@@ -204,18 +204,21 @@ def get_exposure_standard_deviation(entity: Union[RiskFactor, AlternativeRiskFac
...
+    valid_ages = utilities.get_exposure_and_restriction_ages(exposure, entity)
+
     data.drop('age_group_id', axis=1, inplace=True)
     df = []
-    for age_id in set(extract.extract_data(entity, 'exposure', location_id).age_group_id):
+    for age_id in valid_ages:
         copied = data.copy()
         copied['age_group_id'] = age_id
         df.append(copied)
     data = pd.concat(df)
```
halfhorst: concats with no reset usually scare me. Does the data out of extract_data have an index set? I know we didn't reset before, but should we ?

kfwilson: the data out of extract_data does not have an index set but I could reset_index just to be safe?

PR #176, opened 2019-02-11
apply cause restrictions to sequela

michaella21:


michaella21:
Good catch! yes, I updated for those two also!

PR #184, opened 2019-02-13
Feature/sim validator docs

kfwilson:
round 1 of sim validator docs. this is doc strings for all the individual entity-measure validators + return type annotations for everything. working now on doc string-ing all the utilities/helper methods

### DIFF
--- a/src/vivarium_inputs/validation/sim.py
+++ b/src/vivarium_inputs/validation/sim.py
@@ -4,7 +4,6 @@ import numpy as np
 import pandas as pd
 
 from gbd_mapping import ModelableEntity, Cause, Sequela, RiskFactor, CoverageGap, Etiology, Covariate
-
 from vivarium_inputs import utilities, utility_data
 from vivarium_inputs.globals import DataTransformationError, Population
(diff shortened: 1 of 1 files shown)


src/vivarium_inputs/validation/sim.py
```diff
@@ -245,7 +389,41 @@ def validate_excess_mortality(data: pd.DataFrame, entity: Cause, context: Simula
 
 
 def validate_exposure(data: pd.DataFrame, entity: Union[RiskFactor, CoverageGap, AlternativeRiskFactor],
-                      context: SimulationValidationContext):
+                      context: SimulationValidationContext) -> None:
+    """Check the standard set of validations on simulation-prepped exposure
+    data, with the upper boundary of values determined by the distribution type.
+    The broadest age range determined by yll and yld age restrictions is used
+    in the restriction check. Because exposure is filled with two values outside
```
collijk: "Because exposure ..." is only true if you have a dichotomous risk, right?

kfwilson: oh yes. updating that now

PR #193, opened 2019-02-26
Bugfix/multi build artifact

kfwilson:
3 changes in this PR:

1) We weren't using the default output root for multi-build artifact so the individual location artifacts were disappearing and the resulting aggregated artifact only had the metadata keys

2) Add logging to aggregation - this is where most of the diff is coming from. I pulled out several helper functions into utilities so I could use them in both cli.py and aggregation.py to set up logging. This also involved updating a few tests.

3) Bump memory requests for aggregation job. With 3 locations, I as at 33GB and I bumped the memory to 20 slots/35GB so any more than 2 or 3 locations are probably going to get killed on new cluster/run away with memory on old but until we rewrite vivarium inputs to use indices/update aggregation to append to an hdf instead of holding things in memory, we're going to have memory issues.

### DIFF
--- a/src/vivarium_inputs/data_artifact/cli.py
+++ b/src/vivarium_inputs/data_artifact/cli.py
@@ -89,14 +89,16 @@ def multi_build_artifact(model_specification, locations, project, output_root, a
     """
 
     if len(set(locations)) > 15:
-        raise ValueError(f'We can make an artifact for upto 15 locations. You provide {len(set(locations))} locations')
+        raise ValueError(f'We can make an artifact for up to 15 locations. '
+                         f'You provided {len(set(locations))} locations.')
 
     if len(set(locations)) < 1:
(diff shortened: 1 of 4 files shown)


src/vivarium_inputs/data_artifact/cli.py
```diff
@@ -89,19 +89,21 @@ def multi_build_artifact(model_specification, locations, project, output_root, a
     """
 
     if len(set(locations)) > 15:
-        raise ValueError(f'We can make an artifact for upto 15 locations. You provide {len(set(locations))} locations')
+        raise ValueError(f'We can make an artifact for up to 15 locations. '
```
collijk: Drop this to 5

kfwilson: done

PR #196, opened 2019-03-01
add integration yaml specs and a dummy loader component

halfhorst:
good yamls for pulling masses of data


tests/integration/test_cause.yaml
```diff
@@ -0,0 +1,55 @@
+plugins:
+        optional:
+                data:
+                        controller: "vivarium_inputs.data_artifact.ArtifactBuilder"
```
kfwilson: just to make sure I understand how this works - we're using vi ArtifactBuilder here instead of the standard vph ArtifactManager so we can do simulate run and it will build an artifact?

kfwilson:
it won't let me comment on bin files but I think an hdf snuck in here

tests/integration/test_cause.yaml
michaella21: That was an initial thought. For the previous purpose, I actually changed to passthrough during testing for 2017 data, but it might be better to keep as an ArtifactBuilder, since now we want to see that connection also.

kfwilson: Hmm but isn’t this essentially introducing a dependency on vivarium inputs into vivarium public health?

halfhorst:
> it won't let me comment on bin files but I think an hdf snuck in here

whew yeah. I nabbed it

tests/integration/test_cause.yaml
halfhorst: I mean, we could commit it with VPH and leave it up to the person testing to change it. The nature of this test/tool is a little weird. It's kind of just convenience to have this simple component and configuration here

collijk: @kfwilson This test stuff is in vivarium inputs so the dependencies should be okay.

collijk:
We'll want a job on jenkins that runs these test specs at least weekly (but might as well do them daily).  Not for this sprint, but make a ticket for it.

tests/integration/test_cause.yaml
kfwilson: wait I'm still confused about why this is vivarium_inputs.data_artifact.ArtifactBuilder instead of the standard passthrough or vivarium_public_health.dataset_manager.ArtifactManager. In order to do anything with this, I either need to call build_artifact (which will overwite to be the ArtifactBuilder anyways) or change it to the passthrough as far as I can tell. I can't do simulate run with these artifacts because they need to have artifact paths

PR #197, opened 2019-03-03
Shim and sort

halfhorst:
This is the shim to set index as well as the adjustment to sort.

Restriction dicts don't get shimmed, obviously, but one dataframe doesn't as well  - demographic dimensions. That frame has no 'value', nothing apart from our index columns.

I tested this by pulling data and also building artifacts for the smoke tests. Running on the integration yamls now (takin a while)

EDIT: shim everything now!


src/vivarium_inputs/utilities.py
```diff
@@ -175,20 +175,16 @@ def reshape(data: pd.DataFrame, value_cols: List = DRAW_COLUMNS, var_name: str =
...
-    key_cols.extend(other_cols)
-    data = data.sort_values(key_cols).reset_index(drop=True)
-
-    sorted_cols = key_cols
-    if 'value' in data.columns:
-        sorted_cols += ['value']
-    data = data[sorted_cols]
+def sort_hierarchical_data(data: pd.DataFrame) -> pd.DataFrame:
+    """Reorder index labels of a hierarchical index and sort in level order."""
+    sort_order = ['draw', 'location', 'sex', 'age_group_start', 'age_group_end', 'year_start', 'year_end']
```
collijk: I think we want draws to be the last level

collijk: Ah, nevermind.  We have two use cases. One where draws first makes sense and one where draws last makes sense.  Do you know if there's a performance difference in filtered reading from an hdf based on the index level order?

halfhorst: I'm not 100%  but I am pretty sure they are each stored as separate indices so there is no hit. If you poke around HDFStore and look at metadata you can see indices specified as what look like separate structures

collijk: Yeah, I actually don't care about this.  First or last are both okay choices.

src/vivarium_inputs/interface.py
```diff
@@ -143,8 +160,8 @@ def get_demographic_dimensions(location: str) -> pd.DataFrame:
     data = core.get_data(pop, 'demographic_dimensions', location)
     data = utilities.scrub_gbd_conventions(data, location)
     validation.validate_for_simulation(data, pop, 'demographic_dimensions', location)
-    return utilities.sort_data(data)
 
+    return data
```
kfwilson: why aren't we setting index/sorting for demographic dimensions? Because there's no value column? Do we then have to treat this as a special case later on when we read in/use data?

halfhorst: it breaks. Setting it works fine but then it is empty and it causes issues later, beginning with writing the thing to the hdf. It does need to be handled later. If we reset all dataframes then demog dimensions will end up with an extra Unnamed: 0 column. I get around that by checking for a multindex before resetting

kfwilson: ah ok makes sense

src/vivarium_inputs/utilities.py
```diff
@@ -175,20 +176,26 @@ def reshape(data: pd.DataFrame, value_cols: List = DRAW_COLUMNS, var_name: str =
     return data
 
 
+def sort_hierarchical_data(data: pd.DataFrame) -> pd.DataFrame:
+    """Reorder index labels of a hierarchical index and sort in level order."""
+    # sort_order = ['draw', 'location', 'sex', 'age_group_start', 'age_group_end', 'year_start', 'year_end']
```
kfwilson: you don't need this comment anymore since you have the global

PR #202, opened 2019-03-06
Feature/reset indices

halfhorst:
This reset indices just before data are passed on to the simulation from load().

I have to check for dataframe because some data are dictionaries, and I have to check multindex because one dataframe does not have an index set -- demog dimensions.

This requires an accompanying change to vph to fix the function that writes to the artifact.

I tested this by building artifacts for all the smoke tests (Builder) and inspecting their keys, and running them with simulate (Passthrough)

EDIT: We no longer check for multindex because everything comes back with an index set -- see VPH. Tested again by building artifacts for integration yamls and running smoke tests with simulate

### DIFF
--- a/src/vivarium_inputs/data_artifact/builder.py
+++ b/src/vivarium_inputs/data_artifact/builder.py
@@ -85,6 +85,9 @@ class ArtifactBuilder:
         if entity_key not in self.artifact:
             self.process(entity_key)
         data = self.artifact.load(entity_key)
+        # pop structure is a df but not multi-indexed because it has no columns beyond our indexing dimensions
+        if isinstance(data, pd.DataFrame) and isinstance(data.index, pd.MultiIndex):
+            data = data.reset_index()
         return filter_data(data, **__) if isinstance(data, pd.DataFrame) else data
 
     def end_processing(self, event) -> None:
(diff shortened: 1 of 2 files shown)


src/vivarium_inputs/data_artifact/builder.py
```diff
@@ -85,6 +85,9 @@ def load(self, entity_key: str, **__) -> Any:
         if entity_key not in self.artifact:
             self.process(entity_key)
         data = self.artifact.load(entity_key)
+        # pop structure is a df but not multi-indexed because it has no columns beyond our indexing dimensions
```
halfhorst: this is the wrong data I will update this comment here and below

src/vivarium_inputs/data_artifact/builder.py
```diff
@@ -85,6 +85,9 @@ def load(self, entity_key: str, **__) -> Any:
         if entity_key not in self.artifact:
             self.process(entity_key)
         data = self.artifact.load(entity_key)
+        # demog dimensions is a df but not multi-indexed because it has no columns beyond our indexing dimensions
```
kfwilson: instead of having to  handle this special case for demog dimensions in multiple places, why don't we just set an index there too?

halfhorst: commented on the other PR

PR #205, opened 2019-03-11
add set order of levels to reshape + always return dataframe + handle…

kfwilson:
updates to reshape: 
1. add set order of levels (will be important for when we start multiplying/adding/dividing multiple dfs in core)
2. always return a dataframe
3. handle the case where data comes in with partial index set (this is primarily to handle the case of cause_specific_mortality where I have to drop draws out of the index of deaths in order to combine with pop which doesn't have draws and I want to make sure it gets put back in via reshape)

There are also 2 small updates to core because we need to reset indices after groupbys or things go into reshape with a multiindex that contains columns that are also in the data

### DIFF
--- a/src/vivarium_inputs/core.py
+++ b/src/vivarium_inputs/core.py
@@ -296,7 +296,9 @@ def get_relative_risk(entity: Union[RiskFactor, CoverageGap], location_id: int) 
     result = []
     for affected_entity in data.affected_entity.unique():
         df = data[data.affected_entity == affected_entity]
-        df = df.groupby('parameter').apply(lambda d: utilities.normalize(d, fill_value=1))
+        df = (df.groupby('parameter')
+              .apply(utilities.normalize, fill_value=1)
+              .reset_index(drop=True))
         result.append(df)
     data = pd.concat(result)
     data = data.filter(DEMOGRAPHIC_COLUMNS + ['affected_entity', 'affected_measure', 'parameter'] + DRAW_COLUMNS)
@@ -358,7 +360,9 @@ def get_population_attributable_fraction(entity: Union[RiskFactor, Etiology], lo
     data = utilities.convert_affected_entity(data, 'cause_id')
     data.loc[data['measure_id'] == MEASURES['YLLs'], 'affected_measure'] = 'excess_mortality'
     data.loc[data['measure_id'] == MEASURES['YLDs'], 'affected_measure'] = 'incidence_rate'
-    data = data.groupby(['affected_entity', 'affected_measure']).apply(utilities.normalize, fill_value=0)
+    data = (data.groupby(['affected_entity', 'affected_measure'])
+            .apply(utilities.normalize, fill_value=0)
+            .reset_index(drop=True))
     data = data.filter(DEMOGRAPHIC_COLUMNS + ['affected_entity', 'affected_measure'] + DRAW_COLUMNS)
     return data
 
--- a/src/vivarium_inputs/utilities.py
+++ b/src/vivarium_inputs/utilities.py
@@ -9,6 +9,7 @@ import pandas as pd
 from vivarium_inputs import utility_data
 from vivarium_inputs.globals import DRAW_COLUMNS, DEMOGRAPHIC_COLUMNS, SEXES, SPECIAL_AGES
 
+INDEX_COLUMNS = DEMOGRAPHIC_COLUMNS + ['draw', 'affected_entity', 'affected_measure', 'parameter']
 
 ##################################################
 # Functions to remove GBD conventions from data. #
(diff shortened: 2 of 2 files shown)


src/vivarium_inputs/core.py
```diff
@@ -358,7 +360,9 @@ def get_population_attributable_fraction(entity: Union[RiskFactor, Etiology], lo
     data = utilities.convert_affected_entity(data, 'cause_id')
     data.loc[data['measure_id'] == MEASURES['YLLs'], 'affected_measure'] = 'excess_mortality'
     data.loc[data['measure_id'] == MEASURES['YLDs'], 'affected_measure'] = 'incidence_rate'
-    data = data.groupby(['affected_entity', 'affected_measure']).apply(utilities.normalize, fill_value=0)
+    data = (data.groupby(['affected_entity', 'affected_measure'])
+            .apply(utilities.normalize, fill_value=0)
+            .reset_index(drop=True))
```
collijk: Won't this drop your groupby columns?

src/vivarium_inputs/utilities.py
```diff
@@ -163,18 +164,26 @@ def normalize_age(data: pd.DataFrame, fill_value: Real, cols_to_fill: List[str])
...
 
+def get_ordered_index_cols(data_columns: List):
+    return [i for i in INDEX_COLUMNS if i in data_columns] + list(set(data_columns).difference(INDEX_COLUMNS))
+
+
 def reshape(data: pd.DataFrame, value_cols: List = DRAW_COLUMNS, var_name: str = 'draw') -> pd.DataFrame:
     if isinstance(data, pd.DataFrame) and not isinstance(data.index, pd.MultiIndex):
         if set(data.columns).intersection(value_cols):  # reshape wide to long over value_cols
-            data = data.set_index(list(data.columns.difference(value_cols)))
+            data = data.set_index(get_ordered_index_cols(list(data.columns.difference(value_cols))))
```
collijk: You don't need to coerce `data.columns.difference(...)` to a list here.  And, in fact, it disagrees with your call below in the input data type it provides to get_ordered_index_cols.

src/vivarium_inputs/utilities.py
```diff
@@ -163,18 +164,26 @@ def normalize_age(data: pd.DataFrame, fill_value: Real, cols_to_fill: List[str])
     return data
 
 
+def get_ordered_index_cols(data_columns: List):
+    return [i for i in INDEX_COLUMNS if i in data_columns] + list(set(data_columns).difference(INDEX_COLUMNS))
+
+
 def reshape(data: pd.DataFrame, value_cols: List = DRAW_COLUMNS, var_name: str = 'draw') -> pd.DataFrame:
```
collijk: Do we have any tests backing reshape?  It's pretty complicated.

src/vivarium_inputs/core.py
kfwilson: I was finding that affected_entity and affected_measure were still in the data and without the reset_index call, I was ending up with a multiindex that included affected_entity, affected_measure, and the previous numeric index which caused havoc in reshape. I tested these changes by running all the integration yamls and I didn't fail any of the sim validators which check that all the expected columns are there

src/vivarium_inputs/utilities.py
kfwilson: we don't have tests for anything in vivarium_inputs except the validation functions. I think we're somewhat protected because we have the sim validators that check that all the right columns are there and that the right values are in those columns but I can spend some time writing tests for reshape before pushing the indices further?

src/vivarium_inputs/utilities.py
kfwilson: I'm not sure I understand this. If I don't coerce data.columns.difference(...), it ends up being an index? I thought I want to have a list in get_ordered_index_cols

src/vivarium_inputs/core.py
collijk: Weird.  Normalize must do something funny.

src/vivarium_inputs/utilities.py
collijk: I think an index is okay, otherwise line 182 wouldn't work, right?

src/vivarium_inputs/utilities.py
collijk: Let's make a ticket for it.  I'm fine if the integration tests pass for now.

kfwilson: made ticket

src/vivarium_inputs/core.py
kfwilson: I think it may be in the index setting it does in normalize ages but I didn't dig too closely because this worked and didn't involve changing normalize ages which I was afraid would have consequences for other things

src/vivarium_inputs/utilities.py
kfwilson: huh yeah I guess it does. I honestly just forgot the list in line 182 but you're right - I just tested it and it works. getting rid of list now

PR #206, opened 2019-03-12
Feature/only set indices in reshape

kfwilson:
this pushes the reset_index call out to interface + updates the internal measure core functions to use indices

### DIFF
--- a/src/vivarium_inputs/core.py
+++ b/src/vivarium_inputs/core.py
@@ -121,7 +121,7 @@ def get_birth_prevalence(entity: Union[Cause, Sequela], location_id: int) -> pd.
 def get_disability_weight(entity: Union[Cause, Sequela], location_id: int) -> pd.DataFrame:
     if entity.kind == 'cause':
         data = utility_data.get_demographic_dimensions(location_id, draws=True, value=0.0)
-        data = data.set_index(utilities.get_ordered_index_cols(list(data.columns.difference({'value'}))))
+        data = data.set_index(utilities.get_ordered_index_cols(data.columns.difference({'value'})))
         if entity.sequelae:
             for sequela in entity.sequelae:
                 try:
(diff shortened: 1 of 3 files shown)


src/vivarium_inputs/core.py
```diff
@@ -91,11 +89,11 @@ def get_raw_incidence(entity: Union[Cause, Sequela], location_id: int) -> pd.Dat
 
 
 def get_incidence(entity: Union[Cause, Sequela], location_id: int) -> pd.DataFrame:
-    data = get_data(entity, 'raw_incidence', location_id).set_index(DEMOGRAPHIC_COLUMNS + ['draw'])
-    prevalence = get_data(entity, 'prevalence', location_id).set_index(DEMOGRAPHIC_COLUMNS + ['draw'])
+    data = get_data(entity, 'raw_incidence', location_id)
```
collijk: This all cleans up super nicely!

src/vivarium_inputs/utilities.py
```diff
@@ -181,8 +181,8 @@ def reshape(data: pd.DataFrame, value_cols: List = DRAW_COLUMNS, var_name: str =
         else:  # already in right shape so set index
             data = data.set_index(get_ordered_index_cols(data.columns.difference({'value'})))
     elif not data.columns.difference({'value'}).empty:  # we missed some columns that need to be in index
-        data = data.set_index(data.columns.difference({'value'}), append=True)
-        data = data.reorder_levels(get_ordered_index_cols(data.index.names))
+        data = data.set_index(list(data.columns.difference({'value'})), append=True)
+        data = data.reorder_levels(get_ordered_index_cols(set(data.index.names)))
```
collijk: Do you end up with duplicates in your index?

kfwilson: sneaking this into this PR because I shouldn't have pulled it out of the last one - I got overeager removing lists :see_no_evil:

kfwilson: no. the get_ordered_index_cols though uses a difference call that works on indices but data.index.names is a FrozenList so I coerce it to a set so it still works

PR #207, opened 2019-03-12
Feature/scrub gbd with indices

kfwilson:


src/vivarium_inputs/utilities.py
```diff
@@ -26,43 +26,47 @@ def scrub_gbd_conventions(data, location):
 
 
 def scrub_location(data, location):
-    if 'location_id' in data.columns:
-        data = data.drop('location_id', 'columns')
-    data['location'] = location
+    if 'location_id' in data.index.names:
+        data.index = data.index.droplevel(['location_id'])
+    data = pd.concat([data], keys=[location], names=['location'])
```
halfhorst: we want location to no longer be an index ?

src/vivarium_inputs/utilities.py
```diff
@@ -26,43 +26,47 @@ def scrub_gbd_conventions(data, location):
...
     return data
 
 
 def scrub_sex(data):
-    if 'sex_id' in data.columns:
-        data['sex'] = data['sex_id'].map({1: 'Male', 2: 'Female'})
-        data = data.drop('sex_id', 'columns')
+    if 'sex_id' in data.index.names:
+        levels = list(data.index.levels[data.index.names.index('sex_id')])
+        levels = list(map(lambda x: {1: 'Male', 2: 'Female'}.get(x, x), levels))
```
halfhorst: I think you could call map directly on the index and then rename it

src/vivarium_inputs/utilities.py
kfwilson: It is. This is just the easiest way I found to add a new level that is always the same value to the index

src/vivarium_inputs/utilities.py
kfwilson: So what I wanted to do at first was just replace the levels [1, 2] with ['Male', 'Female'] and rename. But we actually don't have a guarantee until we go through the sim validators that there will be just [1, 2] in the levels and you get a really confusing error if you try to replace levels of length != 2 with levels of length 2. I was trying to have a solution that would map 1 -> Male and 2 -> Female but leave everything else the same so the error would come in the sim validators check

src/vivarium_inputs/utilities.py
kfwilson: What I wanted to do first was just replace the levels of location_id to be the location name and then rename the index if location_id was in there because then you only have to mess around with the pd.concat stuff if location is not already in the data (so basically just for life expectancy I think) but again I hit the issue that technically I don't have a guarantee that there is only 1 location id in the data post-core transforms

src/vivarium_inputs/utilities.py
halfhorst: I mean you don't need to coerce to list so you can call map and then list again to force the generator into a list. You can just .map() what data.index.levels gives you and assign that back

halfhorst: like level_i = data.index.names.index('sex_id')
mapped_level = data.index.levels[level_i].map(lambda x: {1: 'Male', 2: 'Female'}.get(x, x))

kfwilson: oh right. huh I swear I tried that and got some FrozenList error but I just tried it again and it works so I must have done something wrong. Changing now.

src/vivarium_inputs/utilities.py
```diff
@@ -26,43 +26,47 @@ def scrub_gbd_conventions(data, location):
...
+    data = pd.concat([data], keys=[location], names=['location'])
     return data
 
 
 def scrub_sex(data):
-    if 'sex_id' in data.columns:
-        data['sex'] = data['sex_id'].map({1: 'Male', 2: 'Female'})
-        data = data.drop('sex_id', 'columns')
+    if 'sex_id' in data.index.names:
+        levels = list(data.index.levels[data.index.names.index('sex_id')])
```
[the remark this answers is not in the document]
collijk: This would make nans for unmapped levels.

src/vivarium_inputs/utilities.py
```diff
@@ -26,43 +26,45 @@ def scrub_gbd_conventions(data, location):
...
 
 
 def scrub_affected_entity(data):
     CAUSE_BY_ID = {c.gbd_id: c for c in causes}
     # RISK_BY_ID = {r.gbd_id: r for r in risk_factors}
-    if 'cause_id' in data.columns:
-        data['affected_entity'] = data.cause_id.apply(lambda cause_id: CAUSE_BY_ID[cause_id].name)
-        data.drop('cause_id', axis=1, inplace=True)
+    if 'cause_id' in data.index.names:
+        levels = list(data.index.levels[data.index.names.index('cause_id')])
```
halfhorst: you've got one down here, too :smile_cat:

src/vivarium_inputs/utilities.py
```diff
@@ -26,43 +26,48 @@ def scrub_gbd_conventions(data, location):
 
 
 def scrub_location(data, location):
-    if 'location_id' in data.columns:
-        data = data.drop('location_id', 'columns')
-    data['location'] = location
+    if 'location_id' in data.index.names and len(data.index.unique('location_id')) == 1:
```
kfwilson: @collijk I can't have non-unique levels so if I have multiple location_ids, I have to just drop location_id and add location as a new index

src/vivarium_inputs/utilities.py
kfwilson: oh good catch. removing now

src/vivarium_inputs/utilities.py
collijk: If your else branch is a generic solution, then let's use that.

src/vivarium_inputs/utilities.py
```diff
@@ -26,43 +26,48 @@ def scrub_gbd_conventions(data, location):
...
+        levels = list(data.index.unique('sex_id').map(lambda x: {1: 'Male', 2: 'Female'}.get(x, x)))
+        data.index = data.index.rename('sex', 'sex_id').set_levels(levels, 'sex')
     return data
 
 
 def scrub_age(data):
-    if 'age_group_id' in data.columns:
+    if 'age_group_id' in data.index.names:
         age_bins = (utility_data.get_age_bins()
                     .filter(['age_group_id', 'age_group_start', 'age_group_end'])
```
collijk: Don't think you actually need the filter here.

kfwilson: I think utility_data.get_age_bins also gives back age_group_name which was maybe why it was originally put in but yeah it shouldn't matter if that is in there too since we map to specific columns. pulled out

src/vivarium_inputs/utilities.py
kfwilson: alright going back to that

PR #208, opened 2019-03-14
Feature/sim validators with indices

kfwilson:
This updates all the the utilities to use indices + updates the tests correspondingly

The changes in the measure validators are 1) switch to lists of index names to pass to the check index/col names functions and 2) switching all filtering to use query to work with indices

### DIFF
--- a/src/vivarium_inputs/validation/sim.py
+++ b/src/vivarium_inputs/validation/sim.py
@@ -457,7 +457,7 @@ def validate_exposure(data: pd.DataFrame, entity: Union[RiskFactor, CoverageGap,
     is_categorical = entity.distribution in ['dichotomous', 'ordered_polytomous', 'unordered_polytomous']
 
     if is_continuous:
-        if set(data.parameter) != {"continuous"}:
+        if set(data.index.unqiue('paramter')) != {"continuous"}:
             raise DataTransformationError("Continuous exposure data should contain "
                                           "'continuous' in the parameter column.")
         valid_kwd = 'continuous'
--- a/src/vivarium_inputs/validation/sim.py
+++ b/src/vivarium_inputs/validation/sim.py
@@ -1124,9 +1124,11 @@ def validate_age_columns(data: pd.DataFrame, context: SimulationValidationContex
 
     """
     expected_ages = (context['age_bins'].filter(['age_group_start', 'age_group_end'])
-                     .set_index(['age_group_start', 'age_group_end'])).index
+                     .set_index(['age_group_start', 'age_group_end'])
+                     .index.reorder_levels(['age_group_start', 'age_group_end']))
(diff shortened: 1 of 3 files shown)


src/vivarium_inputs/validation/sim.py
```diff
@@ -645,11 +647,11 @@ def validate_relative_risk(data: pd.DataFrame, entity: Union[RiskFactor, Coverag
 
     if is_categorical:
         tmrel_cat = sorted(list(entity.categories.to_dict()), key=lambda x: int(x[3:]))[-1]  # chop 'cat' and sort
-        if not (data.loc[data.parameter == tmrel_cat, 'value'] == 1.0).all():
+        if not (data.query('parameter == @tmrel_cat').value == 1.0).all():
```
collijk: I'm super not a fan of the ``@`` stuff in query strings.  It makes most static analysis tools (including the ones in pycharm) super unhappy.  Think you can just use an f-string in this case.

src/vivarium_inputs/validation/sim.py
```diff
@@ -682,14 +684,14 @@ def validate_population_attributable_fraction(data: pd.DataFrame, entity: Union[
...
+    expected_index_names = SCRUBBED_DEMOGRAPHIC_COLUMNS + ['affected_entity', 'affected_measure', 'draw']
+    validate_expected_index_columns(expected_index_names, data.index.names, ['value'], data.columns)
 
     risk_relationship = data.groupby(['affected_entity', 'affected_measure'])
     risk_relationship.apply(validate_standard_columns, context)
 
-    protective_causes = PROTECTIVE_CAUSE_RISK_PAIRS[entity.name] if entity.name in PROTECTIVE_CAUSE_RISK_PAIRS else []
-    protective = data[data.affected_entity.isin([c.name for c in protective_causes])]
+    protective_causes = [c.name for c in PROTECTIVE_CAUSE_RISK_PAIRS.get(entity.name, [])]
+    protective = data.query('affected_entity in @protective_causes')
```
collijk: I'd go back to the ``isin`` stuff here.

src/vivarium_inputs/validation/sim.py
```diff
@@ -971,11 +967,16 @@ def validate_demographic_dimensions(data: pd.DataFrame, entity: Population,
 # UTILITIES #
 #############
 
-def validate_expected_columns(expected_cols: List, existing_cols: List) -> None:
-    """Verify that the passed lists of columns match.
+def validate_expected_index_columns(expected_index_names: List, existing_index_names: List, expected_cols: List,
```
collijk: ``validate_expected_index_and_columns`` is a better name.

src/vivarium_inputs/validation/sim.py
collijk: Oh, may not work with indices.

PR #213, opened 2019-03-29
Rollback/artifact index

collijk:


src/vivarium_inputs/utilities.py
```diff
@@ -185,18 +185,23 @@ def reshape(data: pd.DataFrame, value_cols: List = DRAW_COLUMNS, var_name: str =
...
+    key_cols.extend([c for c in ['location', 'sex', 'age_group_start',
+                                 'age_group_end', 'year_start', 'year_end'] if c in data.columns])
+    other_cols = data.columns.difference(key_cols + ['value'])
+    key_cols.extend(other_cols)
+    data = data.sort_values(key_cols).reset_index(drop=True)
+
+    sorted_cols = key_cols
+    if 'value' in data.columns:
+        sorted_cols += ['value']
+    data = data[sorted_cols]
```
kfwilson: was this filtering here before? key_cols + ['value'] should be the entire set of data columns so this doesn't seem like it will do anything?

src/vivarium_inputs/validation/sim.py
```diff
@@ -157,9 +155,6 @@ def validate_incidence(data: pd.DataFrame, entity: Union[Cause, Sequela], contex
         expected boundary values.
 
     """
-    expected_index_names = SCRUBBED_DEMOGRAPHIC_COLUMNS + ['draw']
```
kfwilson: I think this rolled back too far. We still want to have the check that the expected columns and only the expected columns are present in the data. ``validate_expected_index_and_columns`` used to be just ``validate_expected_columns`` and I think that's what we want here. Same for all functions below

src/vivarium_inputs/validation/sim.py
```diff
@@ -746,22 +708,18 @@ def validate_estimate(data: pd.DataFrame, entity: Covariate, context: Simulation
         group.
 
     """
-    expected_index_names = ['location', 'year_start', 'year_end', 'parameter']
-    if entity.by_sex:
-        expected_index_names += ['sex']
-    if entity.by_age:
-        expected_index_names += ['age_group_start', 'age_group_end']
-    validate_expected_index_and_columns(expected_index_names, data.index.names, ['value'], data.columns)
```
kfwilson: this block (lines 749-754) should all be kept with index swapped for columns - we still want to check columns

src/vivarium_inputs/validation/sim.py
```diff
@@ -893,11 +843,12 @@ def validate_theoretical_minimum_risk_life_expectancy(data: pd.DataFrame, entity
         monotonically decreasing over age.
 
     """
-    expected_index_names = ['age_group_start', 'age_group_end']
-    validate_expected_index_and_columns(expected_index_names, data.index.names, ['value'], data.columns)
+    if 'age_group_start' not in data.columns or 'age_group_end' not in data.columns:
```
kfwilson: this also should just be ``validate_expected_columns``

src/vivarium_inputs/validation/sim.py
```diff
@@ -958,48 +906,13 @@ def validate_demographic_dimensions(data: pd.DataFrame, entity: Population,
...
-    validate_expected_index_and_columns(SCRUBBED_DEMOGRAPHIC_COLUMNS, data.index.names, [], data.columns)
-
     validate_demographic_columns(data, context)
 
 
 #############
 # UTILITIES #
 #############
 
-def validate_expected_index_and_columns(expected_index_names: List, existing_index_names: List, expected_cols: List,
```
kfwilson: There's a version of this that is just ``validate_expected_columns`` and I think we still want that

src/vivarium_inputs/validation/sim.py
```diff
@@ -1059,15 +972,20 @@ def validate_draw_column(data: pd.DataFrame) -> None:
     Raises
     ------
     DataTransformationError
-        If 'draw' column does not contain all values in the range [0, 999].
+         If data does not contain a column named 'draw' or that column does not
+         contain all values in the range [0, 999].
 
     """
-    if set(data.index.unique('draw')) != set(range(1000)):
+    if 'draw' not in data.columns:
```
kfwilson: When I added the ``validate_expected_columns`` method I pulled all of these ``if <col_X> not in data.columns`` blocks out of the ``validate_<col_x>_column`` functions so that there was one place that column names were checked and these individual column validators just checked the set of values in the column. I think we want to go back to that

src/vivarium_inputs/validation/sim.py
```diff
@@ -1059,15 +972,20 @@ def validate_draw_column(data: pd.DataFrame) -> None:
     Raises
     ------
     DataTransformationError
-        If 'draw' column does not contain all values in the range [0, 999].
```
kfwilson: The changes to the docstrings for all these ``validate_<col_X>_column`` shouldn't be rolled back - they should just say the function checks the column values - see the comment below

tests/validation/test_sim.py
```diff
@@ -65,59 +78,76 @@ def test_validate_sex_column_pass():
...
-    ages = (mock_validation_context['age_bins']
-            .filter(['age_group_start', 'age_group_end']))
-    ages = ages.drop(2).set_index(['age_group_start', 'age_group_end'])
+    ages = mock_validation_context['age_bins'].filter(['age_group_start', 'age_group_end'])
+    ages.drop(2, inplace=True)
     with pytest.raises(DataTransformationError):
         sim.validate_age_columns(ages, mock_validation_context)
 
 
+@pytest.mark.parametrize("columns", (('age_group_start',), ('age_group_end',), ('age_group_id_start', 'age_group_end')),
```
kfwilson: This also was rolled back too far. I removed all these missing_column tests when I added the ``validate_expected_columns`` method because that was actually catching the missing columns

tests/validation/test_sim.py
```diff
@@ -185,37 +218,6 @@ def test_check_sex_restrictions(values, restrictions, fill):
     ((1, 2, 2, 2), (False, True), 1.0)
 ], ids=('male', 'female', 'nonzero_fill'))
 def test_check_sex_restrictions_fail(values, restrictions, fill):
-    df = pd.DataFrame({'sex': ['Male', 'Male', 'Female', 'Female'], 'other': [0]*4, 'value': values})
-    df = df.set_index(['sex', 'other'])
+    df = pd.DataFrame({'sex': ['Male', 'Male', 'Female', 'Female'], 'value': values})
     with pytest.raises(DataTransformationError):
         sim.check_sex_restrictions(df, restrictions[0], restrictions[1], fill)
-
```
kfwilson: There's a version of these two tests that's for ``validate_expected_columns``

src/vivarium_inputs/utilities.py
```diff
@@ -185,18 +185,23 @@ def reshape(data: pd.DataFrame, value_cols: List = DRAW_COLUMNS, var_name: str =
         data = data.reorder_levels(get_ordered_index_cols(set(data.index.names)))
     else:  # we've already set the full index
         pass
-    return data
```
kfwilson: we need to return data from reshape. But this version of reshape sets indices so we need to reset them somewhere. Based on what you rolled back in core, I think the right place should probably be at the end of ``core.get_data`` because it looks like where you're using data returned from get_data in other functions in core, you're assuming you don't have indices set

PR #214, opened 2019-04-03
Feature/lbwsg

kfwilson:
here's most of the fixes for lbwsg data including:

1. don't check age/sex ids/restrictions by category in the raw validator for exposure. this was causing problems for some countries where for a certain category, one sex would have all 0s and so trigger that the data was empty but the entity didn't have sex restrictions. I don't really think it makes sense to be checking these by category anyways - if any category has invalid sex/age ids, we'll still raise an error if we check all the data at once.

2. update the check mortality/morbidity flags function to only break if data violates a restriction, not if it doesn't meet one. This is because the RR data for lbwsg is always mortality = 1, morbidity = 0 but the affected causes aren't yll only.

3. the PAF data for lbwsg and neonatal preterm is all 1.0 but it includes age groups that are out of the age restrictions of neonatal preterm. this is an error in the data so I just special-cased it to skip this check in check_paf_rr_exposure_groups

The remaining fixes will be around trying to handle the size of the RR data and will probably have to do something with the indices (although I'm going to have to get clever with intervals it looks like).

### DIFF
--- a/src/vivarium_inputs/core.py
+++ b/src/vivarium_inputs/core.py
@@ -191,8 +191,8 @@ def get_exposure(entity: Union[RiskFactor, AlternativeRiskFactor, CoverageGap], 
     data = extract.extract_data(entity, 'exposure', location_id)
     data = data.drop('modelable_entity_id', 'columns')
 
-    if entity in EXTRA_RESIDUAL_CATEGORY:
-        data = data[data.parameter != EXTRA_RESIDUAL_CATEGORY[entity]]
+    if entity.name in EXTRA_RESIDUAL_CATEGORY:
+        data = data[data.parameter != EXTRA_RESIDUAL_CATEGORY[entity.name]]
 
     if entity.kind in ['risk_factor', 'alternative_risk_factor']:
         data = utilities.filter_data_by_restrictions(data, entity,
--- a/src/vivarium_inputs/core.py
+++ b/src/vivarium_inputs/core.py
@@ -8,7 +8,7 @@ import numpy as np
 
 from vivarium_inputs import utilities, extract, utility_data
 from vivarium_inputs.globals import (InvalidQueryError, DEMOGRAPHIC_COLUMNS, MEASURES, SEXES, DRAW_COLUMNS,
-                                     Population, DataDoesNotExistError)
+                                     Population, DataDoesNotExistError, EXTRA_RESIDUAL_CATEGORY)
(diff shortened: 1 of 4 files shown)


src/vivarium_inputs/validation/raw.py
```diff
@@ -1659,7 +1654,9 @@ def check_paf_rr_exposure_age_groups(paf: pd.DataFrame, context: RawValidationCo
         extra_paf = set(paf.age_group_id).intersection(valid_but_no_rr)
 
         measure = 'YLLs' if measure_id == MEASURES['YLLs'] else 'YLDs'
-        if not_valid_paf:
+        # FIXME: LBWSG paf has data outside neonatal preterm birth age restrictions (but is all 1.0) - K.W. 4/2/19
+        if not_valid_paf and not (entity.name == 'low_birth_weight_and_short_gestation'
```
collijk: Pull this special case into a variable in the globals.

src/vivarium_inputs/core.py
```diff
@@ -191,6 +191,11 @@ def get_exposure(entity: Union[RiskFactor, AlternativeRiskFactor, CoverageGap],
     data = extract.extract_data(entity, 'exposure', location_id)
     data = data.drop('modelable_entity_id', 'columns')
 
+    if entity.name == 'low_birth_weight_and_short_gestation':
```
collijk: Put into a variable in globals.

src/vivarium_inputs/validation/raw.py
```diff
@@ -720,20 +720,18 @@ def validate_exposure(data: pd.DataFrame, entity: Union[RiskFactor, CoverageGap,
     check_years(data, context, 'either')
     check_location(data, context)
 
-    cats = data.groupby('parameter')
```
halfhorst: I think this was discussed while i was gone -- why do we no longer check within categories?

kfwilson: see 1 in the main comment for this PR but basically we can have a situation where for a certain parameter, values could be all 0 for an age/sex group (e.g., all males are 0 for one cat for lbwsg in CIV) so checking age/sex ids/restrictions by sex doesn't work.

halfhorst: oh duh I'm sorry, I forgot to go back and look at your description after reading through. Thanks for including it

PR #220, opened 2019-04-22
Feature/interval indices

kfwilson:
this is a cleaned up version of the previous work I did to add intervals to the indices. I want these changes in first so I can build off them for the wide on draws things without having the first PR for that be huge. I tested this building an artifact - it's slow but it works.

### DIFF
--- a/src/vivarium_inputs/data_artifact/builder.py
+++ b/src/vivarium_inputs/data_artifact/builder.py
@@ -11,8 +11,7 @@ from vivarium_public_health.dataset_manager import (EntityKey, Artifact, get_loc
 from vivarium_public_health.disease import DiseaseModel
 
 from vivarium_inputs.data_artifact.loaders import loader
-from vivarium_inputs.data_artifact.utilities import get_versions
-from vivarium_inputs.utilities import split_interval
+from vivarium_inputs.data_artifact.utilities import get_versions, split_interval
 
 
 _log = logging.getLogger(__name__)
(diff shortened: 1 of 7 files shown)


src/vivarium_inputs/data_artifact/builder.py
```diff
@@ -101,6 +101,9 @@ def process(self, entity_key: EntityKey) -> None:
 def _worker(entity_key: EntityKey, location: str, modeled_causes: Collection[str],
             artifact: Artifact) -> None:
     data = loader(entity_key, location, modeled_causes, all_measures=False)
+    # FIXME: This is a hack since hdf files can't handle pandas.Interval objects
+    data = split_interval('age', 'age_group', data)
```
collijk: Reorder the arguments here and use keywords to clarify what's happening.  E.g. Calls should look like
```
data = split_interval(data, interval_column='age', split_column_prefix='age_group')
```

src/vivarium_inputs/utilities.py
```diff
@@ -42,18 +42,18 @@ def scrub_sex(data):
 def scrub_age(data):
     if 'age_group_id' in data.index.names:
         age_bins = utility_data.get_age_bins().set_index('age_group_id')
-        starts = list(data.index.levels[data.index.names.index('age_group_id')].map(age_bins['age_group_start']))
-        data = (data.assign(age_group_end=(data.index.get_level_values('age_group_id')
-                                           .map(age_bins['age_group_end'])))
-                .set_index('age_group_end', append=True))
-        data.index = data.index.rename('age_group_start', 'age_group_id').set_levels(starts, 'age_group_start')
+        id_levels = data.index.levels[data.index.names.index('age_group_id')]
+        interval_levels = [pd.Interval(age_bins.age_group_start[id],
```
collijk: ``id`` is a builtin you should avoid using as a variable name.

kfwilson: changed

src/vivarium_inputs/data_artifact/builder.py
kfwilson: done

PR #221, opened 2019-04-23
Feature/wide on draws core

kfwilson:
here's the changes that need to happen in core/reshape to stay wide on draws

### DIFF
--- a/src/vivarium_inputs/core.py
+++ b/src/vivarium_inputs/core.py
@@ -124,7 +124,7 @@ def get_birth_prevalence(entity: Union[Cause, Sequela], location_id: int) -> pd.
 def get_disability_weight(entity: Union[Cause, Sequela], location_id: int) -> pd.DataFrame:
     if entity.kind == 'cause':
         data = utility_data.get_demographic_dimensions(location_id, draws=True, value=0.0)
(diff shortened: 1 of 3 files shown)


src/vivarium_inputs/core.py
```diff
@@ -61,14 +61,13 @@ def get_data(entity, measure: str, location: Union[str, int]):
     location_id = utility_data.get_location_id(location) if isinstance(location, str) else location
     data = handler(entity, location_id)
 
-    if measure == 'estimate':
-        value_cols, var_name = COVARIATE_VALUE_COLUMNS, 'parameter'
-    elif measure == 'exposure_distribution_weights':
-        value_cols, var_name = DISTRIBUTION_COLUMNS, 'parameter'
+    if measure in ['structure', 'theoretical_minimum_risk_life_expectancy',
+                   'estimate', 'exposure_distribution_weights']:
```
collijk: Don't covariate estimates still have mean, min, and max?

kfwilson: I just made this change because when I was starting on the sim validators I realized I think we still want to reshape covariate estimates and exposure distribution weights to be long instead of wide or we're going to have to go redo things in vph

PR #224, opened 2019-04-24
rename draw col to value

kfwilson:


src/vivarium_inputs/data_artifact/builder.py
```diff
@@ -87,6 +87,9 @@ def load(self, entity_key: str, **__) -> Any:
         data = self.artifact.load(entity_key)
         if isinstance(data, pd.DataFrame):  # could be metadata dict
             data = data.reset_index()
+            draw_col = [c for c in data if 'draw' in c]
+            if draw_col:
+                data = data.rename(columns={draw_col[0]: 'value'})
```
halfhorst: we are guaranteed a single draw column because that's all you will accept, right? You could also do a c.startswith check, which is ever so slightly a tighter check

kfwilson: No so we aren't actually guaranteed a single draw column by anything in the artifact because we want to allow e.g. `draw in [1, 12, 497]` for people using artifacts interactively. But because we manually set the filter terms for draw in the artifact manager/artifact builder, we know that it will only ever be `draw == builder.configuration.input_data.draw_number`

kfwilson: And I can do a c.startswith check but technically I thought we made no guarantees unless things went through the sim validators so I should be able to assume there are no other potentially confusing columns containing 'draw' that will get mixed in?

halfhorst: seems reasonable

PR #237, opened 2019-07-12
Feature/age restriction ranges

kfwilson:
switches mismatched yll/yld age ranges to warning instead of error + adds special case maximums for subarachnoid hemorrhage excess mortality


src/vivarium_inputs/validation/raw.py
```diff
@@ -1436,17 +1436,12 @@ def check_cause_age_restrictions_sets(entity: Cause) -> None:
     """Check that a cause does not have an age range based on its YLL
     restrictions that is broader than that based on its YLD restrictions.
 
+    Warns users if a wider age range is found so they can further investigate.
```
collijk: You can use a Warns section just like Raises

kfwilson: Nice. Changed!

PR #238, opened 2019-08-06
add information on when get_measure or get_raw_data is more appropriate

KjellSwedin:


docs/source/tutorials/pulling_data.rst
```diff
@@ -352,6 +352,31 @@ argument to fill the location dimension.
     4    Kenya  Female              0.0       0.019178        1994      1995
 
 
+get_measure() versus get_raw_data() -- Which should I use?
```
kfwilson: I like having this which to use section separately but I think the below sub-sections for get_raw_data and get_measure should be incorporated into the existing heading sections for each of those functions (`Pulling Raw GBD Data` and `Pulling Simulation-Prepped Data`). Each of those headings has a sentence or two that partly overlaps with these sub-sections so I think they could be combined to be fleshed out with what you have in these sub-sections.

kfwilson: I also think this `which should I use` section should go at the end of the introductory `Pulling Data` section. I feel like the specific sections on how to pull raw data/simulation-prepped data go into details and telling people up front which one they should spend time reading depending on what their needs are might be better

KjellSwedin: PR comments addressed and some additional rearranging done.

docs/source/tutorials/pulling_data.rst
```diff
@@ -2,174 +2,47 @@
...
+    :depth: 2
+    :local:
+    :backlinks: none
 
 
+Which should I use... get_measure() versus get_raw_data()
+----------------------------------------------------------
+Prefer :func:`vivarium_inputs.interface.get_measure` over
+:func:`vivarium_inputs.interface.get_raw_data`. :func:`vivarium_inputs.interface.get_measure`
+will produce simulation-ready data. If :func:`vivarium_inputs.interface.get_measure`
```
halfhorst: The only augmentation I'd recommend is either matching the language here "simulation-ready data" exactly to the header "simulation-prepped data" so that it is obvious which detailed section will explain how to do this, or turning that phrase into a link to the section. And likewise for the raw data

PR #241, opened 2019-08-28
WIP: Handle tables versioning issues at artifact build time

halfhorst:
The goal is to gracefully handle the decompression errors that come from having the wrong tables version. The general strategy is as follows:

1. Create a get_measure command line function that passes arguments through to get_measure in vivarium_inputs and returns serialized data.
2. Wrap calls to get_measure that occur in the loaders used by the artifact builder with a function that catches the decompression error. On except, the get_measure cli endpoint is called from an environment with updated tables. The data is deserialized and returned

Adding james for now so he can look while this is WIP.

I think this should remain "off" by default. Instead, we should log when we hit a decompression error so it is clear what data caused it. Since it relies on a custom environment with editable installs, I think we should consciously "turn it on" so that we think about what environment we are using.


collijk:
solved by systematically processing the lbwsg results to dodge this issue.

PR #247, opened 2019-09-24
Feature/build artifact slash and burn

kfwilson:
3 changes in this PR:
1. I killed all the multi-loc artifact stuff because it didn't work very well and no one was using it. I instead reverted `multi_build_artifact` back to what it originally was when it was useful: a parallel build artifact for multiple locations at once 
2. I axed all the old cluster stuff since we don't need it anymore
3. I decided while I was at it to refine the version comparison because I got burned so many times by patch version changes. Now it only rebuilds the artifact from scratch if there are any major/minor version differences but not patch differences


collijk:
Will you make a ticket to use the same launch mechanism we use for psimulate (array job using drmaa)?

kfwilson:
> Will you make a ticket to use the same launch mechanism we use for psimulate (array job using drmaa)?

I ended up making a ticket for this instead of doing it because it was more complicated than I originally realized. When I've done this in research repos, I've known what the set of locations was & had an easy way for each task in the array to choose one location via the model specs. Here where the locations are specified as args to the cli, we need to save them somewhere/do something clever so each task can access them. I also realized that there are a lot of utilities in vivarium_cluster_tools we would want to use so probably want to rearrange things. all that to say, it looked like it was going to turn into a bigger task than I have time for

PR #248, opened 2019-10-02
first steps integrating service into vivarium_inputs

KjellSwedin:


src/vivarium_inputs/gbd_service.py
```diff
@@ -0,0 +1,69 @@
...
+from loguru import logger
+import pandas as pd
+from pathlib import Path
+import requests as req
+from typing import List, Callable, TypeVar
+from urllib.parse import urlencode
+
+
+SERVICE_VERSION = '1'
+BASE_URL = f'http://microsim-rancher-p01.hosts.ihme.washington.edu:5000/v{SERVICE_VERSION}'
```
collijk: @halfhorst why doesn't the url map to a specific port?  It seems super clunky to need to specify a port here.

KjellSwedin: Not sure I totally follow... I figured the url or format of the endpoints might change as we moved forward.

collijk: I'm thinking about DNS stuff, which we must not have configured yet.  Rancher gives us hostname stuff (rather than raw ip addrs) for free it seems, but if we want to have a URL that is not machine:port specific, we'll likely have to work that out later.

src/vivarium_inputs/gbd_service.py
```diff
@@ -0,0 +1,69 @@
...
+
+
+def _dataframe_from_response(resp) -> pd.DataFrame:
+    if resp.ok:
+        # return pickle.loads(gzip.decompress(resp.content))
+        return resp.content.decode("utf-8")
+        # return pd.read_csv(io.BytesIO(gzip.decompress(resp.content)), index_col=0)
+        # return pd.read_json(resp.content)
+    else:
+        logger.error(f'GbdServiceError: http response code {resp}')
```
collijk: You can just raise the exception here and the logger will dump the stack trace along with the error message.

src/vivarium_inputs/gbd_service.py
```diff
@@ -0,0 +1,69 @@
...
+        url = f'{url}?{params}'
+    return url
+
+
+def _validate_args(gbd_id, name, location_id, location, source):
+    pass
+
+
+def _make_request(url: str, data_handler: Callable[..., T]) -> T:
+    return data_handler(req.get(url))
```
collijk: I think it's probably wise to split the request handling from the deserialization.  E.g. have this function make the get call and check the response for a 404 or other error code and then return the serialized code.
Have the caller (get_incidence_rate here) then call deserialize separately on the valid raw bytes from the response.  

Keeping I/O separate from transformations makes the code a lot easier to debug and expand.

src/vivarium_inputs/gbd_service.py
halfhorst: we have to know the port to connect, the thing is HTTP defaults to 80 (HTTPS 443) so you never see it for web traffic. For a bunch of services coexisting they need to fan out over different ports. There is maybe a service discovery solution for that but that's beyond me. 

We could set our flask app to listen on 80 and I think everything would work under the default. That's fine as long as we know we won't conflict with anything (we won't). I think it's pretty normal to go with the app default, 5000 for flask in this case

src/vivarium_inputs/extract.py
```diff
@@ -1,15 +1,25 @@
...
-import vivarium_inputs.validation.raw as validation
 
 
-def extract_data(entity, measure: str, location_id: int, validate: bool = True) -> Union[pd.Series, pd.DataFrame]:
+ENDPOINT_DRAWS = 'draws'
+ENDPOINT_SUMMARY = 'summary'
+ENDPOINT_METADATA = 'metadata'
+
+
+def extract_data(entity: ModelableEntity, measure: str, location_id: int,
```
collijk: extraction_map = {'measure/endpoint': 
                                        {'kind_map': 
                                                    {'entity_kind': ('default_source', 'data_type')
                                          'validation_data': [measures]}
}

extraction_map = {'incidence_rate': 
                                        {'kind_map': 
                                                    {'cause': ('como', ENDPOINT_DRAWS),
                                                      'sequela': ('como', ENDPOINT_DRAW),}
                                          'validation_data': ['age_bins']}
}

src/vivarium_inputs/extract.py
```diff
@@ -90,115 +100,331 @@ def extract_data(entity, measure: str, location_id: int, validate: bool = True)
...
+                               "location": None}))
+    resp = make_request(url)
+    check_response(resp)
+    return dataframe_from_response(resp)
+
+
+def extract_exposure_standard_deviation(entity: ModelableEntity, location_id: int) -> pd.DataFrame:
     if entity.kind == 'risk_factor' and entity.name in OTHER_MEID:
-        data = gbd.get_modelable_entity_draws(OTHER_MEID[entity.name], location_id)
+        data = get_exposure(OTHER_MEID[entity.name], location_id)
```
collijk: Test case

PR #258, opened 2020-01-10
update index names used to sort hierarchical data

halfhorst:
We changed the naming conventions around columns/indices in data but didn't update our sorting routine


halfhorst:
I think this should be hotfixed into master as well

PR #259, opened 2020-02-12
use loguru instead of warnings

rmudambi:


tests/conftest.py
```diff
@@ -1,8 +1,21 @@
 import pytest
+import logging
+from _pytest.logging import caplog as _caplog
+from loguru import logger
 
 from gbd_mapping import causes, risk_factors, etiologies
 
 
+@pytest.fixture
```
rmudambi: Added per recommendation here (https://loguru.readthedocs.io/en/stable/resources/migration.html#making-things-work-with-pytest-and-caplog)

src/vivarium_inputs/validation/raw.py
```diff
@@ -253,13 +253,16 @@ def check_cause_metadata(entity: Cause, measure: str) -> None:
         children = "subcauses" if measure == "deaths" else "sequela"
 
         if consistent is not None and not consistent:
-            warnings.warn(f"{measure.capitalize()} data for cause {entity.name} may not exist for {children} in all "
-                          f"locations. {children.capitalize()} models may not be consistent with models for this cause.")
-
+            logger.warning(
```
halfhorst: minor thing, but you expand parens to their own lines sometimes but not always. Don't know how much we care since we aren't linting right now. Is this a Java style thing ??

rmudambi: Yeah I guess it's a more common Java style. I thought it improved readability in the places I did it, and didn't improve it in the places I didn't.

src/vivarium_inputs/core.py
```diff
@@ -1,6 +1,6 @@
+from loguru import logger
```
collijk: NIT: Put the loguru import with the other third party imports (pandas, gbd_mapping, etc.). See: https://www.python.org/dev/peps/pep-0008/#imports

PR #261, opened 2020-11-17
First gbd update checkin, remove CoverageGap and survey data referenc…

KjellSwedin:
…es. Add new tests (won't run under CI due to how long they take to run)

Changes:

 - remove CoverageGap references
 - modify id types as dictated by changes to gbd_access (covid -> cov_id, meid -> me_id, etc.)
 - remove references to survey data (exposure_exists, exposure_standard_deviation_exists, etc.)
 - remove metadata violation checks
 
Add:
 - quasi-unit-test ./tests/extract/test_extract.py. These tests pull GBD data and some can be quite time consuming.
 
Discussion:
 - We don't want to run the test_extract tests as part of CI. I changed travis-test.sh so that
	the "extract" directory is ignored. Should "extract" be renamed something like "gbd_update"? Is there
	a better way to do this?

### DIFF
--- a/ci/travis-test.sh
+++ b/ci/travis-test.sh
@@ -1,4 +1,4 @@
 #!/bin/bash
 
 source builds/venv/bin/activate
-pytest --ignore=./tests/extract tests/
+pytest tests/
--- a/ci/travis-test.sh
+++ b/ci/travis-test.sh
@@ -1,4 +1,4 @@
 #!/bin/bash
 
 source builds/venv/bin/activate
-pytest tests/
+pytest --ignore=./tests/extract tests/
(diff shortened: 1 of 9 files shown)


mattkappel (approved):

ci/travis-test.sh
```diff
@@ -1,4 +1,4 @@
 #!/bin/bash
 
 source builds/venv/bin/activate
-pytest tests/
```
collijk, line 4: Rather than do this, you can use `pytest.mark.skipif` and check to see if the `gbd` you import from utils is the dummy or the module from gbd_access.  https://docs.pytest.org/en/reorganize-docs/new-docs/user/skipping.html

collijk (approved):
You should do the skip thing, but you can do it in the next PR if you'd like.

PR #264, opened 2020-12-04
Add tests for core.py, several tweaks to get tests to pass, single ch…

KjellSwedin:
…eck for CI environment.

Remove obsolete/not implemented code. Small changes so tests pass. Single check for CI environment.

Is there a better way to skip tests under CI? I tried using a fixture in a local conftest.py but pytest didn't recognize the fixture in the skipif context.


src/vivarium_inputs/core.py
```diff
@@ -134,15 +131,12 @@ def get_disability_weight(entity: Union[Cause, Sequela], location_id: int) -> pd
...
-        if not entity.healthstate.disability_weight_exists:
-            data = utility_data.get_demographic_dimensions(location_id, draws=True, value=0.0)
-        else:
-            data = extract.extract_data(entity, 'disability_weight', location_id)
-            data = utilities.normalize(data)
-            cause = [c for c in causes if c.sequelae and entity in c.sequelae][0]
-            data = utilities.clear_disability_weight_outside_restrictions(data, cause, 0.0,
-                                                                          utility_data.get_age_group_ids())
-            data = data.filter(DEMOGRAPHIC_COLUMNS + DRAW_COLUMNS)
+        data = extract.extract_data(entity, 'disability_weight', location_id)
```
collijk, line 134: I think we still want the logic here (return 0s for a non-existent dw), but we no longer have the metadata to lean on.  A try except would work fine.  Warn with `warnings` if we hit the except branch.

tests/extract/check.py
```diff
@@ -0,0 +1,11 @@
+
+def _running_on_ci():
```
collijk, line 2: Running on ci is not the apt thing here I think.  The important bit is that you have no cluster access.  So maybe `_no_gbd_access`?

tests/extract/check.py
```diff
@@ -0,0 +1,11 @@
...
+    try:
+        from vivarium_inputs.globals import GbdDummy
+        return True
+    except:
```
collijk, line 6: Catch a concrete exception

tests/extract/test_core.py
```diff
@@ -0,0 +1,145 @@
...
+
+from gbd_mapping import causes, risk_factors, covariates, ModelableEntity
+from vivarium_inputs import core, utility_data
+from tests.extract.check import RUNNING_ON_CI
+
+
+pytestmark = pytest.mark.skipif(RUNNING_ON_CI, reason="Don't run these tests on the CI server")
+
+
+def success_expected(entity_name, measure_name, location):
```
collijk, line 15: I don't think it would work for the structure you have, but there is a `pytest.mark.xfail` that I'd typically use to mark failing tests.

tests/extract/test_core.py
```diff
@@ -0,0 +1,145 @@
...
+class MCFlag(IntFlag):
+    INCIDENCE_RATE = 1
+    RAW_INCIDENCE_RATE = 2
+    PREVALENCE = 4
+    BIRTH_PREVALENCE = 8
+    DISABILITY_WEIGHT = 16
+    REMISSION_RATE = 32
+    CAUSE_SPECIFIC_MORTALITY_RATE = 64
+    EXCESS_MORTALITY_RATE = 128
+    # not implemented
```
collijk, line 34: Delete

tests/extract/test_core.py
```diff
@@ -0,0 +1,145 @@
...
+    # CASE_FATALITY_RATE = 256
+    DEATHS = 512
+
+
+entity = [
+    (causes.measles,
+        MCFlag.INCIDENCE_RATE
```
collijk, line 41: This is a totally great use case for a bit field, but bit fields are not really idiomatic in python.  

It's not worth changing the implementation at this point, but leave a comment pointing to, e.g., the wikipedia article.  I'm trying to think ahead to our research engineers.  Most newer engineers I've come across may have seen bit level math in a class and then never really needed to think about it much after that.

tests/extract/test_core.py
```diff
@@ -0,0 +1,145 @@
...
+    ('deaths', MCFlag.DEATHS)]
+locations = ['India']
+
+@pytest.mark.parametrize('entity', entity, ids=lambda x: x[0].name)
+@pytest.mark.parametrize('measure', measures, ids=lambda x: x[0])
+@pytest.mark.parametrize('location', locations)
+def test_core_causelike(entity, measure, location):
+    entity_name, entity_expected_field = entity
+    measure_name, measure_id = measure
+    tester = success_expected if (entity_expected_field & measure_id) else fail_expected
```
collijk, line 84: I think you can do `measure_id in entity_expected_field` here which feels a little cleaner.  

Maybe also change `entity_expected_field` to `entity_expected_measure_ids` and you end up with a statement that reads pretty much like normal english.

collijk (changes requested):
The only major things are the disability weight branch and the bare except clause.  Everything else is just a suggestion.

mattkappel (approved):

src/vivarium_inputs/core.py
KjellSwedin: Take a look at this now. It doesn't use `warnings` on except... but this made sense to me.

tests/extract/check.py
KjellSwedin: Changed.

tests/extract/check.py
KjellSwedin: Done.

tests/extract/test_core.py
KjellSwedin: Done.

tests/extract/test_core.py
KjellSwedin: Comments added.

tests/extract/test_core.py
KjellSwedin: Name changed, it does read better.

PR #268, opened 2020-12-08
Add remaining core tests. Skip utilization_rate until gbd access fix.

KjellSwedin:


tests/extract/test_core.py
```diff
@@ -6,6 +6,7 @@
 
 from gbd_mapping import causes, risk_factors, covariates, ModelableEntity
 from vivarium_inputs import core, utility_data
+from vivarium_inputs.mapping_extension import healthcare_entities
 from tests.extract.check import RUNNING_ON_CI
```
collijk, line 10: Didn't this get renamed?

KjellSwedin: healthcare_entities or RUNNING_ON_CI ? I renamed the internal function for RUNNING_ON_CI but left this as the "reason for no gbd access".

collijk (approved):

PR #282, opened 2022-01-21
Changing max population checks to match world population

albrja:
Changed max population upper bound checks to match world population.

Category: feature
JIRA issue: [MIC-2828](https://jira.ihme.washington.edu/browse/MIC-2828)

Verification and Testing
Ran make_artifacts and successfully created artifacts for Sub-Saharan Africa and South Asia.

### DIFF
--- a/src/vivarium_inputs/validation/raw.py
+++ b/src/vivarium_inputs/validation/raw.py
@@ -11,7 +11,7 @@ from vivarium_inputs import utility_data
 from vivarium_inputs.globals import (DRAW_COLUMNS, DEMOGRAPHIC_COLUMNS, SEXES, SPECIAL_AGES, METRICS, MEASURES,
                                      PROTECTIVE_CAUSE_RISK_PAIRS, DataAbnormalError, InvalidQueryError,
                                      DataDoesNotExistError, Population, PROBLEMATIC_RISKS, PAF_OUTSIDE_AGE_RESTRICTIONS,
-                                     EXCLUDE_ABNORMAL_DATA, RISKS_WITH_NEGATIVE_PAF)
+                                     EXCLUDE_ABNORMAL_DATA, RISKS_WITH_NEGATIVE_PAF, WORLD_POP)
 
 from vivarium_inputs.mapping_extension import AlternativeRiskFactor, HealthcareEntity, HealthTechnology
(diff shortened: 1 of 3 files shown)


src/vivarium_inputs/validation/raw.py
```diff
@@ -27,7 +27,6 @@
 MIN_PROTECTIVE_PAF = -1
 MAX_UTILIZATION = 50
 MAX_LIFE_EXP = 90
-MAX_POP = 145_000_000  # pop data includes both sexes combined at this point
```
collijk, line 30: You should be changing this value to the maximum population in a single age group for location id 1.

src/vivarium_inputs/validation/sim.py
```diff
@@ -29,7 +29,7 @@
 VALID_COST_RANGE = (0, {'healthcare_entity': 30_000, 'health_technology': 50})
 # FIXME: bumping for Mexico (max utilization ~422) 9/12/19 - K.W.
 VALID_UTILIZATION_RANGE = (0, 500)
-VALID_POPULATION_RANGE = (0, 75_000_000)
```
collijk, line 32: This should change to the maximum population for a single age _and_ sex group for location id 1.

hussain-jafari (approved):

collijk (changes requested):
The goal in the check here is to provide the **smallest reasonable upper bound** for validation checks.  

In raw validation, you're checking on a data set that is age-specific, but not sex-specific (inferring from the comment on the original value).  That means you want to use an upper bound that is at least as big as the largest single age-group population for the global location id (say, round up to the nearest million people).  In the sim validation context, you're checking a dataset that is age- and sex-specific, so you want to use something a little larger (again, round up to the nearest million) than the largest single age/sex-group population.  

Using the total world population gives you an upper bound that is 10-40x (estimating) larger than the largest thing you should be checking, which means we'd say a population 10x too high would be fine with us.

src/vivarium_inputs/validation/sim.py
mattkappel: Ok, this is different from what the bug (I) said based on our discussion last week. We can make this change.

mattkappel: And we do want these checks to use different numbers, too?

mattkappel (approved):

PR #300, opened 2022-08-09
Dropping model_version_id column due to bug when pulling exposure data

albrja:
### Description
Calling get_standard_data to pull child stunting exposure was breaking due to the handling of model_version_id in the exposure extractor function.
- *Category*: Bugfix
- *JIRA issue*: [MIC-3316](https://jira.ihme.washington.edu/browse/MIC-3316)

The column model_version_id was added to columns to get dropped from the returned data frame.

### Testing
Successfully wrote artifacts for multiple locations.

### DIFF
--- a/src/vivarium_inputs/core.py
+++ b/src/vivarium_inputs/core.py
@@ -220,7 +220,7 @@ def get_exposure(
     entity: Union[RiskFactor, AlternativeRiskFactor], location_id: int
 ) -> pd.DataFrame:
     data = extract.extract_data(entity, "exposure", location_id)
-    data = data.drop("modelable_entity_id", "columns")
+    data = data.drop(["modelable_entity_id", "model_version_id"], "columns")
 
     if entity.name in EXTRA_RESIDUAL_CATEGORY:
         cat = EXTRA_RESIDUAL_CATEGORY[entity.name]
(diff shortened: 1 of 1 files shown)


src/vivarium_inputs/core.py
```diff
@@ -222,6 +216,10 @@ def get_exposure(
     data = extract.extract_data(entity, "exposure", location_id)
     data = data.drop("modelable_entity_id", "columns")
 
+    # Fixme: why are different level locations carrying this column along?
+    if "model_version_id" in data.columns:
+        data = data.drop("model_version_id", "columns")
+
```
mattkappel, line 222: Would it be reasonable to check if the column has only one value -- i.e., avoid a case where you have multiple `model_version_id`s?

stevebachmeier: I think we need to understand where/why it's getting dropped and consider fixing it at that point (basically you're 'fixme').

Regardless, slightly cleaner if you merge these new lines with the previous drop like

data = data.drop(["modelable_entity_id", "model_version_id"], "columns", errors="ignore")

albrja: We could check if there are multiple "model_version_id" being returned but I am not sure what it means for GBD or if it matters.  I do like the cleaner one liner but am nervous about ignoring errors but adding logging might help future debugging.  

I do think we need to find out what is causing this behavior but it is causing the TMREL category for exposure to not exist later when the groupby call is made.

rmudambi: We should investigate why we're getting different columns at different levels of the location hierarchy. We should ask the GBD team why this is.

PR #307, opened 2023-01-24
Fix PAF validation

hussain-jafari:
### Description
- *Category*: bugfix
- *JIRA issue*: [MIC-3794](https://jira.ihme.washington.edu/browse/MIC-3794)

### Changes and notes
Male_expected and female_expected values were being updated within for loop when they should have remained at the value defined before loop. 

### Testing
Pulled PAF data using interface.get_measure which was previously failing validations as a result of this bug.


albrja (approved):

mattkappel:
Is there a test we can add for this?

rmudambi (approved):

PR #349, opened 2024-06-13
Bugfix/location order

albrja:
### Fixes bug in scrub_location function causes locations to not be mapped to the correct location IDs
- *Category*: Bugfix
- *JIRA issue*: https://jira.ihme.washington.edu/browse/MIC-5107

### Changes and notes
-Fixes bug where locations order were out of sync with data and now maps location names to location IDs.
-Hotfix to pin numpy below 2.0

### Testing
All tests pass

### DIFF
--- a/src/vivarium_inputs/utilities.py
+++ b/src/vivarium_inputs/utilities.py
@@ -41,7 +41,7 @@ def scrub_location(data: pd.DataFrame) -> pd.DataFrame:
             locations, level="location"
         )
     else:
-        data = pd.concat([data], keys=location, names=["location"])
+        data = pd.concat([data], keys=locations, names=["location"])
     return data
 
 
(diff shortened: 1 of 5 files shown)


src/vivarium_inputs/utilities.py
```diff
@@ -35,17 +35,22 @@ def scrub_location(data: pd.DataFrame, location: Union[int, List[str]]) -> pd.Da
...
+        utility_data.get_location_id(loc_name): loc_name for loc_name in location_names
+    }
 
     if "location_id" in data.index.names:
-        data.index = data.index.rename("location", level="location_id").set_levels(
-            location, level="location"
-        )
+        index_cols = data.index.names
+        data = data.reset_index()
+        data["location_id"] = data["location_id"].map(location_dict)
```
stevebachmeier, line 49: This threw me off for a second - it might be cleaner to add a new `location` column and then drop `location_id` instead of chaning the ids in the `location_id` column and then renaming it later?

stevebachmeier: Oh, but then your index_cols below wouldn't be correct. It's fine.

stevebachmeier (approved):

hussain-jafari (approved):

PR #350, opened 2024-06-13
Fix bug with estimation years

zmbc:
### Description
- *Category*: bugfix
- *JIRA issue*: None

### Changes and notes

Corrected validation of the `years` parameter.

### Testing

Have not tested this.

### DIFF
--- a/src/vivarium_inputs/extract.py
+++ b/src/vivarium_inputs/extract.py
@@ -104,14 +104,19 @@ def extract_data(
     validation.check_metadata(entity, measure)
 
     # update year_id value for gbd calls
-    if years == None:  # default to most recent year
+    if years is None:  # default to most recent year
         year_id = gbd.get_most_recent_year()
     elif years == "all":
         year_id = None
     else:
         estimation_years = gbd.get_estimation_years()
-        if years not in estimation_years:
-            raise ValueError(f"years must be in {estimation_years}. You provided {years}.")
+        if not isinstance(years, list):
+            years = [years]
+        not_estimated_years = [y for y in years if y not in estimation_years]
+        if len(not_estimated_years) > 0:
+            raise ValueError(
+                f"years must be in {estimation_years}. You provided {not_estimated_years}."
+            )
         year_id = years
 
     try:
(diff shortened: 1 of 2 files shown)


albrja (approved):

src/vivarium_inputs/extract.py
```diff
@@ -104,14 +104,19 @@ def extract_data(
...
-    if years == None:  # default to most recent year
+    if years is None:  # default to most recent year
         year_id = gbd.get_most_recent_year()
     elif years == "all":
         year_id = None
     else:
         estimation_years = gbd.get_estimation_years()
-        if years not in estimation_years:
-            raise ValueError(f"years must be in {estimation_years}. You provided {years}.")
+        if not isinstance(years, list):
```
rmudambi, line 113: I think this is slightly cleaner
```python
else:
    year_id = years if isinstance(years, list) else [years]
    not_estimated_years = set(year_id).difference(gbd.get_estimation_years())
    if not_estimated_years:
        raise ...
```

src/vivarium_inputs/utility_data.py
```diff
@@ -69,12 +69,19 @@ def get_demographic_dimensions(
...
-    if years == "all":
+    if years is None:
+        years = [gbd.get_most_recent_year()]
+    elif years == "all":
         years = range(min(estimation_years), max(estimation_years) + 1)
-    else:
-        if years and years not in estimation_years:
-            raise ValueError(f"years must be in {estimation_years}. You provided {years}.")
-        years = [years] if years else [gbd.get_most_recent_year()]
+    elif not isinstance(years, list):
```
rmudambi, line 76: Why are the solutions different in these two files? The problem appears to be the same.

zmbc: The solutions are slightly different; one is calling GBD shared functions and the other is creating a data structure from scratch (years = None) won't work in the latter case. But they were more different than they needed to be; I've updated this.

rmudambi (approved):

zmbc:
Superseded by #351.

PR #352, opened 2024-07-10
pull non loglinear rrs

hussain-jafari:
### Description
- *Category*: feature
- *JIRA issue*: https://jira.ihme.washington.edu/browse/MIC-5129

### Changes and notes
Convert exposure column to parameter column and keep exposures as NaNs.
Update allowed upper limit of RR values.
Add RR extraction tests back in to test suite.

### Testing
Ran test suites (except for years='all' which was run in a notebook). Ran pulled data through simulation validations.

### DIFF
--- a/src/vivarium_inputs/extract.py
+++ b/src/vivarium_inputs/extract.py
@@ -1,5 +1,6 @@
 from typing import List, Optional, Union
 
+import numpy as np
 import pandas as pd
 from gbd_mapping import Cause, Covariate, Etiology, ModelableEntity, RiskFactor, Sequela
 
@@ -331,12 +332,10 @@ def extract_relative_risk(
     year_id: Optional[Union[int, str, List[int]]] = None,
 ) -> pd.DataFrame:
     data = gbd.get_relative_risk(entity.gbd_id, location_id, year_id=year_id)
-    # TODO: [MIC-4891] Process new relative risk data format properly
-    if not data["exposure"].isna().all():
-        raise DataAbnormalError(
-            "Relative risk data in new format with 1000 exposure values. Our processing is not "
-            "currently able to process data in this format."
-        )
+    # TODO: new validations?
+    if not data["exposure"].isna().any() and data["parameter"].isna().all():
+        data["parameter"] = data["exposure"]
+        data["exposure"] = np.nan
     data = filter_to_most_detailed_causes(data)
     if entity.gbd_id == 136:  # non-exclusive breastfeeding
         data = data.loc[data["age_group_id"].isin([3, 388])]
--- a/tests/extract/test_core.py
+++ b/tests/extract/test_core.py
@@ -167,11 +167,10 @@ entity_r = [
     ),
 ]
 measures_r = [
-    ("exposure", MRFlag.EXPOSURE),
-    ("exposure_standard_deviation", MRFlag.EXPOSURE_SD),
-    ("exposure_distribution_weights", MRFlag.EXPOSURE_DIST_WEIGHTS),
-    # TODO: Add back in with Mic-4936
-    # ("relative_risk", MRFlag.RELATIVE_RISK),
+    #("exposure", MRFlag.EXPOSURE),
+    #("exposure_standard_deviation", MRFlag.EXPOSURE_SD),
+    #("exposure_distribution_weights", MRFlag.EXPOSURE_DIST_WEIGHTS),
+    ("relative_risk", MRFlag.RELATIVE_RISK),
     ("population_attributable_fraction", MRFlag.PAF),
 ]
 locations_r = ["India"]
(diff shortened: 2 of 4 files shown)


src/vivarium_inputs/extract.py
```diff
@@ -331,12 +332,10 @@ def extract_relative_risk(
     year_id: Optional[Union[int, str, List[int]]] = None,
 ) -> pd.DataFrame:
     data = gbd.get_relative_risk(entity.gbd_id, location_id, year_id=year_id)
-    # TODO: [MIC-4891] Process new relative risk data format properly
-    if not data["exposure"].isna().all():
-        raise DataAbnormalError(
-            "Relative risk data in new format with 1000 exposure values. Our processing is not "
-            "currently able to process data in this format."
-        )
+    # TODO: new validations?
```
albrja, line 335: Remove todo

src/vivarium_inputs/extract.py
```diff
@@ -331,12 +332,10 @@ def extract_relative_risk(
...
+    if not data["exposure"].isna().any() and data["parameter"].isna().all():
```
albrja, line 336: Nit: I find the not with and boolean logic confusing sometimes but you can just do notna() instead of not isna.

albrja: Also do you need to worry about if some of parameter is empty but not all?

stevebachmeier: I agree w/ Jim that `if not data["exposure"].notna().all() and data["parameter"].isna().all():` is clearer

src/vivarium_inputs/extract.py
```diff
@@ -331,12 +332,10 @@ def extract_relative_risk(
...
+        data["parameter"] = data["exposure"]
+        data["exposure"] = np.nan
```
stevebachmeier, line 338: This seems like a big change that applies to things not specific to pulling non-loglinear rrs?

src/vivarium_inputs/extract.py
hussain-jafari: This is a reminder that we should discuss any possible new validations that we might want to perform as a result of this new data structure.

src/vivarium_inputs/extract.py
hussain-jafari: Ah I didn't know about notna I like that more as well. And Jim - I'm not sure since I'm not that familiar with this type of RR data from GBD but you're right in thinking that it's more of a check that there's anything at all in parameter column.

src/vivarium_inputs/extract.py
hussain-jafari: Non log linear RRs are the only type of data where we have data in the exposure column that we've seen so this should only apply to those sort of data.

albrja (approved):

tests/extract/test_core.py
```diff
@@ -194,6 +193,9 @@ def test_core_risklike(entity, measure, location):
 def test_year_id_risklike(entity, measure, location, years):
     entity_name, entity_expected_measure_ids = entity
     measure_name, measure_id = measure
+    # exposure-parametrized RRs for all years requires a lot of time and memory to process
```
stevebachmeier, line 196: But don't we still need to test it? Could we mark that specific combination of parameters as slow and have it run automatically overnight or something like we do elsewhere?

hussain-jafari: Yeah I would definitely prefer to do that. Do you have an example of where we do that - I believe psuedopeople?

aflaxman:
This works!  But it generates a gazillion warnings for me; when I run it with this command, 
```
import vivarium_inputs as vii, gbd_mapping
df = vii.get_measure(gbd_mapping.risk_factors.diet_high_in_sodium, 'relative_risk', 'Washington')
```
I get a lot of stuff like the following
```
2024-07-16 08:52:26.812 | WARNING  | vivarium_inputs.validation.raw:check_age_restrictions:2239 - Data was expected to contain all age groups between ids 8 and 235 but was missing the following: {8, 9}.
```

tests/extract/test_core.py
```diff
@@ -194,12 +194,24 @@ def test_year_id_risklike(entity, measure, location, years):
...
     measure_name, measure_id = measure
     # exposure-parametrized RRs for all years requires a lot of time and memory to process
-    if entity[0].name =='high_systolic_blood_pressure' and measure[0] == 'relative_risk' and years=='all':
-        pass
+    if (
+        entity[0].name == "high_systolic_blood_pressure"
+        and measure[0] == "relative_risk"
+        and years == "all"
+    ):
+        pytest.skip(reason="need --runslow option to run")
```
stevebachmeier, line 202: Ok, I misunderstood your slack message. I thought you were trying to mark as slow here rather than directly mark as skip.

stevebachmeier (approved):

PR #354, opened 2024-07-16
disability weight location processing bugfix

hussain-jafari:
### Description=
- *Category*: bugfix
- *JIRA issue*: https://jira.ihme.washington.edu/browse/MIC-5171

### Changes and notes
Don't duplicate data for locations since this is already done in gbd access.
Only drop year_id column if it is present in data.

### Testing
Pulled disability weight for diarrheal diseases with no year specified, for 2021, for 2019 and 2021, and for all years.

### DIFF
--- a/src/vivarium_inputs/extract.py
+++ b/src/vivarium_inputs/extract.py
@@ -240,9 +240,8 @@ def extract_disability_weight(
             data["key"] = 1
             year_df["key"] = 1
             # Merge to get the Cartesian product, then drop the key column
-            data = pd.merge(
-                data.drop("year_id", axis=1, errors="ignore"), year_df, on="key"
-            ).drop("key", axis=1)
+            data = data.drop("year_id", axis=1) if "year_id" in data.columns else data
+            data = pd.merge(data, year_df, on="key").drop("key", axis=1)
         else:
             data["year_id"] = year_id
     return data
--- a/src/vivarium_inputs/extract.py
+++ b/src/vivarium_inputs/extract.py
@@ -229,18 +229,9 @@ def extract_disability_weight(
         "all",
         location_id,
     )
-    disability_data = disability_weights.loc[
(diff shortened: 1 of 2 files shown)


src/vivarium_inputs/extract.py
```diff
@@ -249,9 +240,9 @@ def extract_disability_weight(
             data["key"] = 1
             year_df["key"] = 1
             # Merge to get the Cartesian product, then drop the key column
-            data = pd.merge(data.drop("year_id", axis=1), year_df, on="key").drop(
-                "key", axis=1
-            )
+            data = pd.merge(
+                data.drop("year_id", axis=1, errors="ignore"), year_df, on="key"
```
hussain-jafari, line 244: Here I'm just adding errors='ignore' so we don't error out if year_id isn't in data (we usually expect it not to be after looking at the disability weight flat file, but I'm just being safe and dropping it if it is present).

src/vivarium_inputs/extract.py
```diff
@@ -229,18 +229,9 @@ def extract_disability_weight(
         "all",
         location_id,
     )
-    disability_data = disability_weights.loc[
+    data = disability_weights.loc[
         disability_weights.healthstate_id == entity.healthstate.gbd_id, :
     ]
-    # Update location_id to match original location id
```
hussain-jafari, line 235: We already do this in gbd access - Zeb was seeing duplicates in our data from these lines

src/vivarium_inputs/extract.py
albrja: Ignoring errors is pretty scary. I think it would be a lot better to do a simple check for this column before merging so you don't inadvertently cover up a different error caused by a data schema issue.

rmudambi (approved):

albrja (approved):

PR #367, opened 2024-10-04
Sbachmei/mic 5317/fully implement incidence rate cause

stevebachmeier:
### Description

- *Category*:  feature
- *JIRA issue*: https://jira.ihme.washington.edu/browse/MIC-5317

### Changes and notes
 This implements db_queries.get_outputs to get mean data for
incidence rate. 

NOTE: this is one of two PRs for this ticket. The next will implement
for a Sequela.

### DIFF
--- a/src/vivarium_inputs/core.py
+++ b/src/vivarium_inputs/core.py
@@ -38,8 +38,7 @@ def get_data(
     measure: str,
     location: str | int | list[str | int],
     years: int | str | list[int] | None,
-    data_type: str | list[str],
-    value_columns: list[str],
+    data_type: utilities.DataType,
 ) -> pd.DataFrame:
     """Pull raw GBD data for measure and entity.
 
@@ -59,11 +58,7 @@ def get_data(
         Years for which to extract data. If None, get most recent year. If 'all',
         get all available data.
     data_type
-        Data type for which to extract data. Supported values include 'mean' for
-        getting mean data and 'draw' for getting draw-level data. Can also be a list
-        of values to get multiple data types.
-    value_columns
-        List of value columns to be pulled.
+        The DataType object for which to extract data.
 
     Returns
     -------
@@ -131,8 +126,8 @@ def get_data(
             utility_data.get_location_id(location) if isinstance(location, str) else location
         ]
 
-    data = handler(entity, location_id, years, data_type, value_columns)
-    data = utilities.reshape(data, value_columns)
+    data = handler(entity, location_id, years, data_type)
+    data = utilities.reshape(data, data_type.value_columns)
 
     return data
 
@@ -141,12 +136,9 @@ def get_raw_incidence_rate(
     entity: Cause | Sequela,
     location_id: list[int],
     years: int | str | list[int] | None,
-    data_type: str | list[str],
-    value_columns: list[str],
+    data_type: utilities.DataType,
 ) -> pd.DataFrame:
-    data = extract.extract_data(
-        entity, "incidence_rate", location_id, years, data_type, value_columns
-    )
+    data = extract.extract_data(entity, "incidence_rate", location_id, years, data_type)
     if entity.kind == "cause":
         restrictions_entity = entity
     else:  # sequela
@@ -156,8 +148,8 @@ def get_raw_incidence_rate(
     data = utilities.filter_data_by_restrictions(
         data, restrictions_entity, "yld", utility_data.get_age_group_ids()
     )
-    data = utilities.normalize(data, value_columns, fill_value=0)
-    data = data.filter(DEMOGRAPHIC_COLUMNS + value_columns)
+    data = utilities.normalize(data, data_type.value_columns, fill_value=0)
+    data = data.filter(DEMOGRAPHIC_COLUMNS + data_type.value_columns)
     return data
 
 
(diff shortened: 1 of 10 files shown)


src/vivarium_inputs/core.py
```diff
@@ -39,6 +39,7 @@ def get_data(
     location: str | int | list[str | int],
     years: int | str | list[int] | None,
     data_type: str | list[str],
+    value_columns: list[str],
```
stevebachmeier, line 42: I was calling the `get_value_columns` in different places so figured I'd just pull the arg up

src/vivarium_inputs/extract.py
```diff
@@ -28,23 +28,13 @@
 )
 
 
-class DataTypeNotImplementedError(NotImplementedError):
```
stevebachmeier, line 31: This was becoming difficult to deal with here so I just moved it into utilities where it's actually used

src/vivarium_inputs/validation/raw.py
```diff
@@ -303,15 +296,16 @@ def check_healthcare_entity_metadata(entity: HealthcareEntity, measure: str) ->
         )
 
 
-#################################################
-#   VALIDATE RAW DATA ENTITY SPECIFIC METHODS   #
-# --------------------------------------------- #
-# Signatures to match wrapper validate_raw_data #
```
stevebachmeier, line 309: This isn't true

src/vivarium_inputs/core.py
rmudambi: Maybe better than this is at the top level (that you need it), convert a `data_type` string into a new `DataType` object that has a `name` and a `value_columns` attribute.

src/vivarium_inputs/interface.py
```diff
@@ -65,10 +65,20 @@ def get_measure(
         Dataframe standardized to the format expected by `vivarium` simulations.
     """
     utilities.validate_data_type(data_type)
-    data = core.get_data(entity, measure, location, years, data_type)
+    if measure in [
```
rmudambi, line 68: I think this is a merge conflict error. You moved this logic inside `get_value_columns()` before.

src/vivarium_inputs/validation/raw.py
```diff
@@ -1999,20 +2041,18 @@ def check_columns(expected_cols: List, existing_cols: List) -> None:
...
-        )
-    elif set(existing_cols) > set(expected_cols):
-        logger.warning(
-            f"Data returned extra columns: {set(existing_cols).difference(set(expected_cols))}."
-        )
+    missing_columns = set(expected_cols).difference(set(existing_cols))
+    extra_columns = set(existing_cols).difference(set(expected_cols))
+    if missing_columns:
+        raise DataAbnormalError(f"Data is missing columns: {missing_columns}.")
+    if extra_columns:
```
rmudambi, line 2048: It is possible both to have missing and extra columns. Maybe we should log the extra columns warning before raising on missing columns, so we get informed of both if both have happened.

rmudambi (approved):

src/vivarium_inputs/interface.py
stevebachmeier: Ack, yeah good catch

tests/test_interface.py
```diff
@@ -152,66 +223,66 @@ def test_get_incidence_rate(
...
+                return_value=get_mocked_location_path_to_global(),
+            )
+            mocker.patch(
+                "vivarium_inputs.utility_data.get_raw_age_bins",
+                return_value=get_mocked_age_bins(),
+            )
+            mocker.patch(
+                "vivarium_inputs.utility_data.get_raw_location_ids",
+                return_value=get_mocked_location_ids(),
+            )
```
albrja, line 267: Nice to all of these

src/vivarium_inputs/utilities.py
```diff
@@ -572,76 +572,96 @@ class DataTypeNotImplementedError(NotImplementedError):
...
+            If a data type is not supported.
+        ValueError
+            If `data_type` is not a string or a list of strings.
+        """
+
+        # Temporarily raise for lists of data types
+        if isinstance(data_type, list):
+            raise DataTypeNotImplementedError("Lists of data types are not yet supported.")
+
+        if not isinstance(data_type, (list, str)):
```
rmudambi, line 614: This check is maybe not necessary, as if this is violated, it should get caught by the set difference check below?

src/vivarium_inputs/validation/sim.py
```diff
@@ -126,12 +126,8 @@ def validate_for_simulation(
...
         Otherwise, validate that data has most recent year.
         Defaults to False.
-    data_type
-        Data type of the extracted data. Supported values include 'mean' for
-        mean data and 'draw' for draw-level data. Can also be a list of values
-        for multiple data types.
-    value_columns
-        List of column names in `data` that contain the values to be validated.
+    data_request
+        DataRequest object of the extracted data.
```
rmudambi, line 130: This should be DataType?

src/vivarium_inputs/utilities.py
```diff
@@ -572,76 +572,96 @@ class DataTypeNotImplementedError(NotImplementedError):
...
-        cols = ["value"]
-    else:
-        data_type_col_mapping = {
-            "mean": MEAN_COLUMNS,
-            "draw": DRAW_COLUMNS,
-        }
-        cols = []
+    def __init__(self, data_type: str | list[str], measure: str | None) -> None:
+        self._validate_data_type(data_type)
+        self.request = data_type
```
rmudambi, line 590: Is `types` a better name for this?

rmudambi (approved):

src/vivarium_inputs/utilities.py
stevebachmeier: I think we still need it. I simply have't implemented a request for, say, "mean" AND "draw" columns. But since those are both valid data types, it wouldn't be caught in the set difference blow. _that's_ looking specifically for unknown datatypes altogether, list or not (e.g. data_type = ["means", "draws", "foo"])

PR #369, opened 2024-10-08
Sbachmei/mic 5317 5392/implement incidence rate for sequela

stevebachmeier:
### Description

- *Category*:  feature
- *JIRA issue*: https://jira.ihme.washington.edu/browse/MIC-5392

### Changes and notes
 

In this case, the changes made on the vivarium_gbd_access side of things
allowed for this to simply work. all the hard work was just setting up for
a new test.

### Testing

all pass

### DIFF
--- a/tests/test_interface.py
+++ b/tests/test_interface.py
@@ -27,6 +27,11 @@ CAUSE = causes.hiv_aids
 LOCATION = 163  # India
 YEAR = 2021
 
+SUPPORTED_DATA_TYPES = {
+    "mean": MEAN_COLUMNS,
+    "draw": DRAW_COLUMNS,
+}
+
 
 def get_mocked_estimation_years() -> list[int]:
     """Mocked estimation years for testing."""
--- a/tests/test_interface.py
+++ b/tests/test_interface.py
@@ -221,7 +221,7 @@ def get_mocked_incidence_rate_get_outputs_common_data() -> pd.DataFrame:
 def mocked_cause_incidence_rate_get_draws() -> pd.DataFrame:
     """Mocked vivarium_gbd_access data for testing."""
 
-    df = get_mocked_incidence_rate_get_draws_common_data().copy()
+    df = get_mocked_incidence_rate_get_draws_common_data()
 
     # Add on specific columns
     df["cause_id"] = int(HIV_AIDS.gbd_id)
@@ -240,7 +240,7 @@ def mocked_cause_incidence_rate_get_draws() -> pd.DataFrame:
 def mocked_sequela_incidence_rate_get_draws() -> pd.DataFrame:
     """Mocked vivarium_gbd_access data for testing."""
 
-    df = get_mocked_incidence_rate_get_draws_common_data().copy()
+    df = get_mocked_incidence_rate_get_draws_common_data()
 
     # Add on specific columns
     df["sequela_id"] = int(HIV_AIDS_DRUG_SUSCEPTIBLE_TB_WO_ANEMIA.gbd_id)
@@ -267,7 +267,7 @@ def mocked_sequela_incidence_rate_get_draws() -> pd.DataFrame:
 def mocked_cause_incidence_rate_get_outputs() -> pd.DataFrame:
     """Mocked vivarium_gbd_access data for testing."""
 
-    df = get_mocked_incidence_rate_get_outputs_common_data().copy()
+    df = get_mocked_incidence_rate_get_outputs_common_data()
 
     # Add on other metadata columns
     df["cause_id"] = int(HIV_AIDS.gbd_id)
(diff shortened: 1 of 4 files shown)


tests/test_interface.py
```diff
@@ -64,6 +67,64 @@ def get_mocked_age_groups() -> dict[int, str]:
     }
 
 
+MOCKED_AGE_GROUP_ENDPOINTS = {
```
rmudambi, line 70: Nit: Would this be easier to read/maintain if you unzipped tuples of `(age_start, age_end)`, i.e. something like this:
```python
AGE_GROUPS = list(zip(*[(0.0, 0.07), (0.07, 0.5), ...]))
MOCKED_AGE_GROUP_ENDPOINTS = {
    key: endpoints for key, endpoints in zip(["age_start", "age_end"], zip(*AGE_GROUPS))
}
```
It would be fewer lines and easier to add/remove age groups, but the code is a bit harder to parse.

tests/test_interface.py
```diff
@@ -64,6 +67,64 @@ def get_mocked_age_groups() -> dict[int, str]:
...
+        0.5,
+        1.0,
+        2.0,
+        5.0,
+        10.0,
+        15.0,
+        20.0,
+        25.0,
+        30.0,
+        0.01917808,
```
rmudambi, line 83: The age start/end here are the same?

tests/test_interface.py
```diff
@@ -104,6 +163,18 @@ def mocked_hiv_aids_incidence_rate_draws() -> pd.DataFrame:
...
         df[col] = np.random.uniform(0, 0.0004, size=len(df))
 
+    return df
+
+
+@pytest.fixture
+def mocked_incidence_rate_cause_draws() -> pd.DataFrame:
+    """Mocked vivarium_gbd_access data for testing."""
+
+    df = get_mocked_common_incidence_rate_draws().copy()
```
rmudambi, line 173: This copy doesn't appear necessary. You're copying a newly created dataframe that isn't being persisted anywhere else. 

Did you instead want `get_mocked_common_incidence_rate_draws()` to be a constant (i.e. `MOCKED_COMMON_INCIDENCE_RATE_DRAWS`)? Then you would need to use `.copy()`

tests/test_interface.py
```diff
@@ -190,15 +336,21 @@ def no_cache(mocker: MockerFixture) -> None:
...
 def test_get_incidence_rate(
+    entity: Cause | Sequela,
     data_type: str | list[str],
     mock_gbd: bool,
-    mocked_hiv_aids_incidence_rate_draws: pd.DataFrame,
-    mocked_hiv_aids_incidence_rate_means: pd.DataFrame,
+    mocked_incidence_rate_cause_draws: pd.DataFrame,
+    mocked_incidence_rate_cause_means: pd.DataFrame,
+    mocked_incidence_rate_sequela_draws: pd.DataFrame,
+    mocked_incidence_rate_sequela_means: pd.DataFrame,
```
rmudambi, line 353: Wouldn't it be better to parametrize the fixture used here as `expected_draws` and `expected_means`, or just have them be constants rather than fixtures? That way you don't need the branching if/else pattern in the test code.

rmudambi: Oh never mind, I see your TODO to do exactly that

tests/test_interface.py
```diff
@@ -267,19 +426,29 @@ def test_get_incidence_rate(
             )
```
rmudambi, line 426: Can you extract this mocking to a private method? Can be done later if you are already planning a clean-up PR.

tests/test_interface.py
```diff
@@ -267,19 +426,29 @@ def test_get_incidence_rate(
...
+        assert all(data.notna())
+        assert all(data >= 0)
+        # Check metadata index values (note that there may be other metadata returned)
+        expected_metadata = {
+            "location": {"India"},
+            "sex": {"Male", "Female"},
+            "age_start": set(MOCKED_AGE_GROUP_ENDPOINTS["age_start"]),
+            "age_end": set(MOCKED_AGE_GROUP_ENDPOINTS["age_end"]),
+            "year_start": {2021},
+            "year_end": {2022},
```
rmudambi, line 451: Shouldn't this also include  `cause_id` or `sequela_id`?

rmudambi (approved):

tests/test_interface.py
stevebachmeier: Oh, yeah that's lingering from when I had it as a constant

tests/test_interface.py
stevebachmeier: Hmm, yeah I like that

stevebachmeier: Actually, it occurs to me that all this data is already in the age groups dataset I've already saved out so I'll just use that

tests/test_interface.py
stevebachmeier: No, once cleaned up that is lost

PR #370, opened 2024-10-09
Sbachmei/clean up test fixtures

stevebachmeier:
### Description

- *Category*:  refactor; test
- *JIRA issue*: na

### Changes and notes
 
I was originally trying my best to keep the tests DRY by using a single
one and parametrizing across everything, but it got pretty unreadable.
This PR cleans it up in three primary ways:
1. Stop parameterizing by "mock_gbd" and instead just split that
logic into two different tests.
2. DRY up after splitting into two tests by creating a common helper/check
function each calls
3. It was ugly passing in all mocked fixtures regardless of which one gets used.
Instead, rely on `request.get_fixture_value` to determine which fixture to use.

### Testing

all pass

### DIFF
--- a/tests/test_interface.py
+++ b/tests/test_interface.py
@@ -23,7 +23,6 @@ from vivarium_inputs.globals import (
 from vivarium_inputs.interface import get_measure
 from vivarium_inputs.utilities import DataTypeNotImplementedError
 
-
 #############
 # CONSTANTS #
 #############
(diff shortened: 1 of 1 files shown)


tests/test_interface.py
```diff
@@ -166,11 +175,53 @@ def get_mocked_common_incidence_rate_draws() -> pd.DataFrame:
     return df
 
 
+def get_mocked_incidence_rate_get_outputs_common_data() -> pd.DataFrame:
```
stevebachmeier, line 178: This is a direct copy/paste to just move the function (except I changed the function name as well)

tests/test_interface.py
```diff
@@ -212,71 +263,36 @@ def mocked_incidence_rate_sequela_draws() -> pd.DataFrame:
     return df
 
 
-def get_mocked_common_incidence_rate_means() -> pd.DataFrame:
```
stevebachmeier, line 215: I just renamed the function and moved it up towards the top of the file

tests/test_interface.py
```diff
@@ -379,76 +426,78 @@ def test_get_incidence_rate(
...
-
-            data = get_measure(**kwargs)
-            mock_extract_incidence_rate.assert_called_once()
-            mock_extract_prevalence.assert_called_once()
-        if not mock_gbd and runslow:
-            # Test actual data retrieval
-            if NO_GBD_ACCESS:
-                pytest.skip("Need GBD access to run this test")
-            data = get_measure(**kwargs)
+        # Use the parameter ID combination to request the correct fixture data
```
stevebachmeier, line 429: I didn't change the logic here; I just removed the `if mock_gbd` check and unindented as appropriate

tests/test_interface.py
```diff
@@ -379,76 +426,78 @@ def test_get_incidence_rate(
...
+        mocker.patch(
+            "vivarium_inputs.utility_data.get_raw_age_bins",
+            return_value=get_mocked_age_bins(),
+        )
+        mocker.patch(
+            "vivarium_inputs.utility_data.get_raw_location_ids",
+            return_value=get_mocked_location_ids(),
+        )
 
-        # Check data
```
stevebachmeier, line 437: All of the data quality assertions have been removed from the (now two) tests and moved into a helper function at the bottom of this file

rmudambi (approved):
You can ignore a few of my comments in the other PR since they're handled here

PR #372, opened 2024-10-17
Sbachmei/mic 5407/prep for e2e tests

stevebachmeier:
### Description

- *Category*:  test
- *JIRA issue*: https://jira.ihme.washington.edu/browse/MIC-5407

### Changes and notes
 There are good tests in the old ones that we have basically never
been running (and are currently "turned off" by previously being moved
into the TESTS_TO_IMPLEMENT folder which pytest doesn't see).

The goal is to turn those back on w/ 
- `data_type` implemented as appropriate
- extra validations
- slow tests run on push and all run on schedule

This PR is just a bit of prep work.


src/vivarium_inputs/extract.py
```diff
@@ -200,7 +200,7 @@ def _get_most_recent_year() -> int:
 #######################
 
 
-def extract_prevalence(
```
stevebachmeier, line 203: I just shuffled the order of these functions (`extract_prevalence and `extract_incidence_rate`) to match the mapping in gbd.core. I didn't touch their logic.

tests/e2e/test_interface.py
```diff
@@ -35,6 +35,88 @@
 YEAR = 2021
 
 
+#########
```
stevebachmeier, line 38: This is a direct copy/paste - I just mmoved the tests to the top and the fixtures/helpers to the bottom

src/vivarium_inputs/core.py
```diff
@@ -271,92 +270,112 @@ def get_disability_weight(
...
 ) -> pd.DataFrame:
-    data = extract.extract_data(
-        entity,
-        "exposure",
-        location_id,
-        validate=True,
-        years=years,
-    )
+
+    if data_type.type != "draws":
```
albrja, line 365: Can you not move this to extract data so you don't have to paste this everywhere?

albrja (approved):

src/vivarium_inputs/core.py
stevebachmeier: I would've had to just paste it everywhere there. But I agree it would've been smarter to do it that way.

tests/conftest.py
```diff
@@ -40,52 +42,14 @@ def emit(self, record):
...
-        causes.ischemic_heart_disease,
-        causes.ischemic_stroke,
-        causes.hemorrhagic_stroke,
-        causes.tetanus,
-        causes.diabetes_mellitus,
-        causes.all_causes,
-    ]
-
-
-@pytest.fixture
```
stevebachmeier, line 56: none of these fixtures are being used

rmudambi (approved):

src/vivarium_inputs/extract.py
```diff
@@ -200,7 +200,7 @@ def _get_most_recent_year() -> int:
...
+def extract_incidence_rate(
```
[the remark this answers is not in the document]
stevebachmeier: Simply so it matches the mapping above. Just an ocd thing (and also helps me keep track of what I'm doing)

PR #374, opened 2024-10-18
Sbachmei/mic 5407 5442/add cause and sequela like tests

stevebachmeier:
### Description

- *Category*:  test
- *JIRA issue*: https://jira.ihme.washington.edu/browse/MIC-5407

### Changes and notes
 
this implements a few tests for Cause and one for Sequela data
retrieval. There are instances where I can't get data (likely due to my
ignorance) and so I'm adding FIXME skips to them to fix in the future.

Following PRs will implement the other types (Covariate, RiskFactor, etc)
as well as mocked versions.

### DIFF
--- a/src/vivarium_inputs/core.py
+++ b/src/vivarium_inputs/core.py
@@ -370,12 +370,12 @@ def get_exposure(
     data = extract.extract_data(entity, "exposure", location_id, years, data_type)
     data = data.drop("modelable_entity_id", "columns")
 
+    value_columns = data_type.value_columns
+
     if entity.name in EXTRA_RESIDUAL_CATEGORY:
         cat = EXTRA_RESIDUAL_CATEGORY[entity.name]
         data = data.drop(labels=data.query("parameter == @cat").index)
-        data[data_type.value_columns] = data[data_type.value_columns].clip(
-            lower=MINIMUM_EXPOSURE_VALUE
-        )
+        data[value_columns] = data[value_columns].clip(lower=MINIMUM_EXPOSURE_VALUE)
 
     if entity.kind in ["risk_factor", "alternative_risk_factor"]:
         data = utilities.filter_data_by_restrictions(
@@ -390,23 +390,23 @@ def get_exposure(
         #  FIXME: We fill 1 as exposure of tmrel category, which is not correct.
         data = pd.concat(
             [
-                utilities.normalize(exposed, data_type.value_columns, fill_value=0),
-                utilities.normalize(unexposed, data_type.value_columns, fill_value=1),
+                utilities.normalize(exposed, value_columns, fill_value=0),
+                utilities.normalize(unexposed, value_columns, fill_value=1),
             ],
             ignore_index=True,
         )
 
         # normalize so all categories sum to 1
-        cols = list(set(data.columns).difference(data_type.value_columns + ["parameter"]))
-        sums = data.groupby(cols)[data_type.value_columns].sum()
+        cols = list(set(data.columns).difference(value_columns + ["parameter"]))
+        sums = data.groupby(cols)[value_columns].sum()
         data = (
             data.groupby("parameter")
-            .apply(lambda df: df.set_index(cols).loc[:, data_type.value_columns].divide(sums))
+            .apply(lambda df: df.set_index(cols).loc[:, value_columns].divide(sums))
             .reset_index()
         )
     else:
-        data = utilities.normalize(data, data_type.value_columns, fill_value=0)
-    data = data.filter(DEMOGRAPHIC_COLUMNS + data_type.value_columns + ["parameter"])
+        data = utilities.normalize(data, value_columns, fill_value=0)
+    data = data.filter(DEMOGRAPHIC_COLUMNS + value_columns + ["parameter"])
     return data
 
 
@@ -508,22 +508,23 @@ def get_relative_risk(
     data.loc[morbidity & ~mortality, "affected_measure"] = "incidence_rate"
     data.loc[~morbidity & mortality, "affected_measure"] = "cause_specific_mortality_rate"
     data = _filter_relative_risk_to_cause_restrictions(data)
+    value_columns = data_type.value_columns
     data = data.filter(
         DEMOGRAPHIC_COLUMNS
         + ["affected_entity", "affected_measure", "parameter"]
-        + data_type.value_columns
+        + value_columns
     )
     data = (
         data.groupby(["affected_entity", "parameter"])
-        .apply(utilities.normalize, cols_to_fill=data_type.value_columns, fill_value=1)
+        .apply(utilities.normalize, cols_to_fill=value_columns, fill_value=1)
         .reset_index(drop=True)
     )
     if entity.distribution in ["dichotomous", "ordered_polytomous", "unordered_polytomous"]:
         tmrel_cat = utility_data.get_tmrel_category(entity)
         tmrel_mask = data.parameter == tmrel_cat
-        data.loc[tmrel_mask, data_type.value_columns] = data.loc[
-            tmrel_mask, data_type.value_columns
-        ].mask(np.isclose(data.loc[tmrel_mask, data_type.value_columns], 1.0), 1.0)
+        data.loc[tmrel_mask, value_columns] = data.loc[tmrel_mask, value_columns].mask(
+            np.isclose(data.loc[tmrel_mask, value_columns], 1.0), 1.0
+        )
     # Coerce location_id from global to requested location - location_id is list of length 1
     data["location_id"] = location_id[0]
 
@@ -542,6 +543,7 @@ def get_population_attributable_fraction(
             f"Data type(s) {data_type.type} are not supported for this function."
         )
 
+    value_columns = data_type.value_columns
     causes_map = {c.gbd_id: c for c in causes}
     if entity.kind == "risk_factor":
         data = extract.extract_data(
@@ -583,13 +585,13 @@ def get_population_attributable_fraction(
         data = utilities.filter_data_by_restrictions(
             data, cause, "inner", utility_data.get_age_group_ids()
         )
-        if np.any(data[data_type.value_columns] < 0):
+        if np.any(data[value_columns] < 0):
             logger.warning(
                 f"{entity.name.capitalize()} has negative values for paf. These will be replaced with 0."
             )
-            other_cols = [c for c in data.columns if c not in data_type.value_columns]
+            other_cols = [c for c in data.columns if c not in value_columns]
             data.set_index(other_cols, inplace=True)
-            data = data.where(data[data_type.value_columns] > 0, 0).reset_index()
+            data = data.where(data[value_columns] > 0, 0).reset_index()
 
     data = utilities.convert_affected_entity(data, "cause_id")
     data.loc[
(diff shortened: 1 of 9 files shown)


src/vivarium_inputs/core.py
```diff
@@ -143,8 +147,8 @@ def get_incidence_rate(
     years: int | str | list[int] | None,
     data_type: utilities.DataType,
 ) -> pd.DataFrame:
-    data = get_data(entity, "raw_incidence_rate", location_id, years, data_type)
-    prevalence = get_data(entity, "prevalence", location_id, years, data_type)
+    data = get_data(entity, "raw_incidence_rate", location_id, years, data_type.type)
```
rmudambi, line 150: In this case (and possibly every case in this file), the data type we already have is exactly the same as the one we want for raw incidence rate even though they have different measures. If we need a different DataType, we could create it here and just pass it to `get_data()`

We don't really use the measure. I think it's simpler and more elegant to modify the `init` of `DataType` like this, and have a docstring that communicates the measures for which you want to force the value column (the parameter name could probably be improved):

```python
def __init__(self, data_type: str | list[str], force_value_column: bool = False) -> None:
    ...
```

src/vivarium_inputs/core.py
```diff
@@ -241,23 +245,23 @@ def get_disability_weight(
...
-        cause_prevalence = get_data(entity, "prevalence", location_id, years, data_type)
+        cause_prevalence = get_data(entity, "prevalence", location_id, years, data_type.type)
         data = (data / cause_prevalence).fillna(0).reset_index()
     else:  # entity.kind == 'sequela'
         try:
             data = extract.extract_data(
                 entity, "disability_weight", location_id, years, data_type
             )
-            data = utilities.normalize(data, data_type.value_columns)
+            data = utilities.normalize(data, DRAW_COLUMNS)
```
rmudambi, line 258: You shouldn't need to change this? Especially if you make the change I suggested above.

src/vivarium_inputs/validation/raw.py
```diff
@@ -523,19 +528,26 @@ def validate_birth_prevalence(
 
     if entity.kind == "cause":
         check_sex_restrictions(
-            data, context, entity.restrictions.male_only, entity.restrictions.female_only
+            data,
+            context,
+            entity.restrictions.male_only,
+            entity.restrictions.female_only,
+            DRAW_COLUMNS,
```
rmudambi, line 535: Why is this not `value_columns`? Same question everywhere you are doing this in `checkXRestrictions()`.

src/vivarium_inputs/validation/raw.py
```diff
@@ -1367,6 +1445,7 @@ def validate_estimate(
         any values in columns do not match the expected set of values.
 
     """
+    # The value columns are irrelevant of the data_type in this case
```
hussain-jafari, line 1448: Should this be "the value columns of the data_type"? It's a bit confusing to parse like this

tests/e2e/test_get_measure.py
```diff
@@ -1,27 +1,48 @@
...
+
+
+def fail_expected(entity, measure, location, data_type, raise_type=Exception):
+    with pytest.raises(raise_type):
+        _df = get_measure(entity, measure, location, years=None, data_type=data_type)
 
 
 ENTITIES_C = [
+    # (entity, applicable_measures)
+    # NOTE: 'raw_incident_rate' and 'deaths' should not be called directly from `get_measure()`
```
hussain-jafari, line 45: nit: raw_incidence_rate

src/vivarium_inputs/validation/raw.py
```diff
@@ -1336,13 +1406,19 @@ def validate_etiology_population_attributable_fraction(
...
 ) -> None:
     raise NotImplementedError()
 
 
 def validate_estimate(
-    data: pd.DataFrame, entity: Covariate, context: RawValidationContext
+    data: pd.DataFrame,
+    entity: Covariate,
+    context: RawValidationContext,
+    value_columns: list[str],
```
rmudambi, line 1421: If the value_columns are never used, why are we passing them to this function?

tests/e2e/test_get_measure.py
```diff
@@ -1,27 +1,48 @@
...
+from vivarium_inputs.utilities import DataTypeNotImplementedError
 
 pytestmark = pytest.mark.skipif(
     NO_GBD_ACCESS, reason="Cannot run these tests without vivarium_gbd_access"
 )
 
+LOCATION = "India"
+
+
+@pytest.fixture(autouse=True)
```
rmudambi, line 22: This is identical to the `no_cache` fixture in GBD Access, right? Is this something we should pull out and put in VTU?

rmudambi (changes requested):

tests/e2e/test_get_measure.py
```diff
@@ -102,59 +191,89 @@ def test_get_measure_causelike(entity_details, measure, location):
...
+####################
+# HELPER FUNCTIONS #
+####################
+
+
+def check_data(data: pd.DataFrame, data_type: str) -> None:
+    """Check the data returned."""
+    assert all(col in data.columns for col in SUPPORTED_DATA_TYPES[data_type])
+    assert all(data.notna())
+    assert all(data >= 0)
```
hussain-jafari, line 257: This isn't always true - for example, PAFs can be negative like for FPG. I asked CC about this and they said "Hi Hussain Jafari this is expected in some risks - for example some risks may have a protective effect and reduce the chance of an outcome occurring." I'm not sure if it happens for any measure other than PAFs though. I believe we are capping our PAFs to be greater than -1 though.

hussain-jafari: I think it's only not true for PAFs though

src/vivarium_inputs/validation/raw.py
stevebachmeier: As discussed, I was having issues since value_columns isn't implemented everywhere. I'll put it back and try again, though.

src/vivarium_inputs/validation/raw.py
stevebachmeier: Because these validation functions get called off of a mapper like `validator["measure"](..., value_columns)`

src/vivarium_inputs/validation/raw.py
stevebachmeier: Most "value_columns" are very specifically defined by the `data_type`, i.e. a data type of "draws" means `value_columns` == ['draw_*'] and a data_type of "means" means `value_columns` == ['val']. But here, I'm trying to convey that regardless of what `data_type` is, we are using custom value_columns.

tests/e2e/test_get_measure.py
stevebachmeier: Yeah I was thinking about that when we were discussing future projects. We could. It would only really be used by these two repos I think but at least would be centralized.

tests/e2e/test_get_measure.py
stevebachmeier: Hmm, good call. I think I'll leave it for now and re-address when something breaks.

tests/e2e/test_get_measure.py
rmudambi: Can you create a JIRA ticket to do this?

rmudambi (approved):

src/vivarium_inputs/validation/raw.py
hussain-jafari: Ah maybe something like "the value columns do not depend on the data_type in this case"?

hussain-jafari (approved):

src/vivarium_inputs/utilities.py
```diff
@@ -573,39 +573,44 @@ class DataTypeNotImplementedError(NotImplementedError):
...
+            self._validate_data_type(data_type)
 
-    def __init__(self, data_type: str | list[str], measure: str | None) -> None:
-        self._validate_data_type(data_type)
         self.type = data_type
-        self.value_columns = self._get_value_columns(data_type, measure)
+        """Data type(s) for which to get value columns. Supported values include 
+        'mean' for getting mean data and 'draw' for getting draw-level data, or 
+        None if you want to specifically use the columns passed in. Can
+        also be a list of values to get multiple data types."""
```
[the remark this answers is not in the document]
stevebachmeier: "or None if you want to specifically use the columns passed in." was my attempt at that. I'll convert to a list to hopefully make it more clear.

I also hope this clarifies that my intent was that a None data type won't go away when means is fully implemented. It's actually reserved more for when you're passing in value_cols, i.e. the type of data really isn't "means" or "draws".

Clearly there's a refactor in order but I haven't thought yet about how to do it

src/vivarium_inputs/utilities.py
```diff
@@ -588,10 +588,20 @@ def __init__(
...
+        -----
+        This is mutually exclusive with `value_cols`, i.e. one of the two is
+        required but not both.
+
+        Supported values include:
+        - 'means' for getting mean data
+        - 'draws' for getting draw-level data
+        - None if the requested data is niche. In this case, the value columns
+        must be passed in directly via the `value_cols` argument.
+        """
```
rmudambi, line 604: I still think `value_cols` (and supporting None as a `data_type` will go away once `get_means` is implemented. Maybe we should talk if you disagree.

src/vivarium_inputs/validation/raw.py
```diff
@@ -476,7 +476,10 @@ def validate_prevalence(
 
 
 def validate_birth_prevalence(
-    data: pd.DataFrame, entity: Union[Cause, Sequela], context: RawValidationContext
+    data: pd.DataFrame,
+    entity: Union[Cause, Sequela],
```
patricktnast, line 480: nit: change to Cause | Sequela

src/vivarium_inputs/validation/sim.py
```diff
@@ -88,7 +88,7 @@ def validate_for_simulation(
     measure: str,
     location: int | str | list[int | str],
     years: int | None,
-    data_request: DataType,
+    data_type: DataType,
```
patricktnast, line 91: It looks like there are several calls to `validate_for_simulation` that don't include a data_type argument.

src/vivarium_inputs/utilities.py
```diff
@@ -573,39 +573,57 @@ class DataTypeNotImplementedError(NotImplementedError):
...
+            raise ValueError("Cannot pass both 'data_type' and 'value_cols'.")
+        if data_type is None and value_cols is None:
+            raise ValueError("Must pass either 'data_type' or 'value_cols'.")
+
+        if data_type is not None:
+            self._validate_data_type(data_type)
 
-    def __init__(self, data_type: str | list[str], measure: str | None) -> None:
-        self._validate_data_type(data_type)
         self.type = data_type
```
patricktnast, line 590: i find this (and subsequent gbd calls arguments given as data_type.type) slightly annoying, but i dont know that theres a more legible alternative.

tests/unit/test_utilities.py
```diff
@@ -45,6 +45,22 @@ def test_normalize_sex_no_sex_id():
...
+    if data_type is None and value_cols is None:
+        with pytest.raises(ValueError, match="Must pass either 'data_type' or 'value_cols'."):
+            utilities.DataType(data_type, value_cols)
+    elif data_type is not None and value_cols is not None:
+        with pytest.raises(
+            ValueError, match="Cannot pass both 'data_type' and 'value_cols'."
+        ):
+            utilities.DataType(data_type, value_cols)
+    else:
+        _data_type = utilities.DataType(data_type, value_cols)
```
patricktnast, line 61: couldn't you just do `utilities.DataType(data_type, value_cols)` here?

albrja (approved):

src/vivarium_inputs/validation/sim.py
stevebachmeier: They just aren't implemented yet. We runtime error higher up the call stack if you pass in anything but draws.

src/vivarium_inputs/utilities.py
stevebachmeier: You and me both, brother.

tests/unit/test_utilities.py
stevebachmeier: As opposed to assigning the return to something? Yeah, that would work. We (and obviously I) are not totally consistent in tests with this if we're not actually doing anything with the return

PR #375, opened 2024-10-21
Sbachmei/mic 5407 5443/mocked data for cause and sequela tests

stevebachmeier:
### Description

- *Category*:  test
- *JIRA issue*: https://jira.ihme.washington.edu/browse/MIC-5443

### Changes and notes
 
This (finally) implements mocked data for the Cause and Sequela tests.

### Testing

mocked tests pass when I run locally. A followup PR will turn on the mocked tests for github actions.


tests/e2e/test_get_measure.py
```diff
@@ -263,18 +316,8 @@ def check_data(data: pd.DataFrame, data_type: str) -> None:
...
     for idx, expected in expected_metadata.items():
         assert set(data.index.get_level_values(idx)) == expected
-
-
-###############
-# MOCKED DATA #
-###############
-
-
-def get_mocked_age_bins() -> pd.DataFrame:
```
stevebachmeier: moved into mocked_gbd.py

tests/e2e/test_get_measure.py
```diff
@@ -88,37 +99,36 @@ def fail_expected(entity, measure, location, data_type, raise_type=Exception):
...
-@pytest.mark.parametrize("entity_details", ENTITIES_C, ids=lambda x: x[0].name)
-@pytest.mark.parametrize("measure", MEASURES_C, ids=lambda x: x)
+@pytest.mark.parametrize("entity_details", CAUSES, ids=lambda x: x[0].name)
+@pytest.mark.parametrize("measure", CAUSE_MEASURES, ids=lambda x: x)
 @pytest.mark.parametrize(
     "data_type", ["means", "draws", ["means", "draws"]], ids=("means", "draws", "means_draws")
 )
-def test_get_measure_causelike(entity_details, measure, data_type):
-
-    entity, entity_expected_measures = entity_details
```
stevebachmeier: lots of this logic is moved into the `run_test` helper function

albrja (approved):
I think this looks pretty good. Two things we need to think about but out of scope for this: 
-Where should the mocked gbd stuff live? I was originally thinking VTU but now maybe it makes more sense to live in gbd_mapping or gbd_access?
-We need to make sure how we mock these data schemas gbd produces is easy to update if/when they change their format and this seems like a great start.

tests/fixture_data/population_india_2021.csv
```diff
@@ -0,0 +1,76 @@
+age_group_id,location_id,year_id,sex_id,population,run_id
+2,163,2021,1,218309.69883222756,359
```
rmudambi, line 2: Can't you make up this mocked data on the fly? We don't actually care what the population values are as long as they are within "reasonable" bounds.

rmudambi (approved):

tests/fixture_data/population_india_2021.csv
stevebachmeier: It's the reasonable bounds that seemed tricky b/c the values vary so much dependent on age group

PR #388, opened 2024-12-20
Sbachmei/mic 5245/handle unused tests

stevebachmeier:
### Description

- *Category*:  test
- *JIRA issue*: https://jira.ihme.washington.edu/browse/MIC-5245

### Changes and notes
 
We previously "turned off" a bunch of core.py and extract.py tests by
moving them into a TESTS_TO_BE_IMPLEMENTED/ directory (which
pytest doesn't see). Most of these imo are not worth the effort to
implement since we've already got their `get_measure()` e2e test
equivalents. However, there are some location- and year-request
tests that are _not_ being tested in the `test_get_measure()` tests.

I moved the location- and year- tests into a new `test_core.py` unit
test module and updated them to work with the mocked_gbd 
fixtures.

NOTE:
It's pretty hacky, but I had to mark two tests as xfail b/c they require
CC's `core-maths` library. Unfortunately, I can't find a clean way
to mark a specific set of test parameters as xfail unless I'm willing
to group them together (I'm not); then I can do [this approach](https://docs.pytest.org/en/latest/how-to/skipping.html#skip-xfail-with-parametrize).

So instead, I'm skipping the two parameters in-test and then I created
a new `test_xfailed()` test that runs just those parameters as xfail.

For some reason they are getting marked as FIXMEs rather than XFAILs, 
but I did confirm that the test will properly fail if, say, the expected error
type changed (or, analogously, the test actually passed)

### Testing

I tested all run successfully in an env that does NOT have vivarium_gbd_access
installed (i.e. whatever needs to be mocked is getting mocked).


tests/unit/test_core.py
```diff
@@ -122,51 +113,95 @@ def test_year_id_causelike(entity_details, measure, location, years, data_type):
...
+        and data_type == "draws"
+    ):
+        pytest.skip("FIXME: [mic-5542] continuous rrs cannot validate")
     entity, entity_expected_measures = entity_details
     if measure in entity_expected_measures:
-        check_year_in_data(entity, measure, location, years, data_type)
+        check_year_in_data(entity, measure, LOCATION, years, data_type, mocker)
+
+
+@pytest.mark.xfail(raises=ModuleNotFoundError, reason="Cannot import core-maths", strict=True)
```
stevebachmeier, line 157: As mentioned in the description, this seems to be behaving as expected except for the fact that the pytest log is showing up as `SKIPPED` / `s` rather than `XFAIL` / `x`. Not sure why.

rmudambi (approved):

patricktnast (approved):

PR #395, opened 2025-01-27
Change PAF affected_measure to CSMR

zmbc:
### Description
- *Category*: bugfix
- *JIRA issue*: None

### Changes and notes

See comment, and let me know if any additional context from [the Slack thread]() would be helpful to include.

### Testing

None, which makes me nervous. How do we "integration test" this with existing models?


src/vivarium_inputs/core.py
```diff
@@ -750,7 +759,11 @@ def _filter_relative_risk_to_cause_restrictions(data: pd.DataFrame) -> pd.DataFr
         cause = causes_map[cause]
         if measure == "cause_specific_mortality_rate":
             start, end = utilities.get_age_group_ids_by_restriction(cause, "yll")
-        else:  # incidence_rate
+        else:
+            assert (
```
rmudambi, line 763: Don't use `assert` statements in code like this. If you want it to fail if the measure is not incidence_rate, throw the appropriate exception manually.

zmbc: Gotcha, any pointers on what the appropriate exception is in a case like this, where if the condition wasn't true it would certainly mean a bug in `vivarium_inputs`, not user error or anything like that?

zmbc: I can also just revert this change if you prefer the comment!

zmbc: Took a pass at this with `DataTransformationError`.

rmudambi (approved):

albrja (approved):

PR #397, opened 2025-01-29
move data deps to core

patricktnast:
### Description

- *Category*: 
- *JIRA issue*: https://jira.ihme.washington.edu/browse/MIC-5673

### Changes and notes
 
Summary. If/when we move automated builds off of github-actions and to jenkins, we can then move `vivarium_gbd_access` and `core-maths` from the `data` setup requirements and into the core ones.
### Testing

this should fail github actions and pass jenkins


albrja (approved):

setup.py
```diff
@@ -24,12 +24,12 @@
         "vivarium>=1.2.1",
         "gbd_mapping>=4.0.0, <5.0.0",
         "loguru",
+        "vivarium-gbd-access>=4.1.0, <5.0.0",
```
albrja, line 27: Won't this break when we try to release it on pypi?

patricktnast: that's a good point, we want to just stop making vivarium inputs a pypi package (and use artifactory instead), but you're correct in pointing out that that blocks this.

